PR #51, opened 2015-08-25
bug 1005013 - add newrelic to server process

groovecoder:
This is how I started it. It should work like this, but some follow-ups could be to see if we can add more granular monitoring/profiling into macros and macro workers?


groovecoder:
Oops, just noticed I need to make the app name come from an environment variable too.

groovecoder:
Updated `newrelic.js` to pull all config from environment variables:
```
NEW_RELIC_APP_NAME
NEW_RELIC_LICENSE_KEY
NEW_RELIC_HIGH_SECURITY
```
It still works. And when I left my local vm running some background render tasks, I'm starting to see some good data about which pages take the longest for KumaScript to execute:
mdn-local-ks-nr

darkwing:
R+.  Builds properly, tests pass, eye-test passes.

PR #55, opened 2015-09-02
bug 1005013 - move newrelic conf to nconf

groovecoder:
Moves the newrelic config values into the same `kumascript_settings_local.json` [`nconf`](https://www.npmjs.com/package/nconf) file that we use for other environment-specific kumascript config values. Also moves the config-loading code (via `nconf`) out of `run.js` and into `lib/kumascript.conf.js` now that we're starting to use the config values in more than just `run.js`.

Spot-check:
1. Check out this branch
2. Go to our [Setup New Relic](https://rpm.newrelic.com/accounts/263620/applications/setup#) page for node, and click "Reveal License Key"
3. Copy the license key value into `kumascript_settings_local.json` in the root `kuma` directory like so:
```
    "log": {
        "console": true,
        "file": {
            "filename": "/home/vagrant/logs/kumascript.log",
            "maxsize": 500000
        }
    },
    "newrelic": {
        "license_key": "..."
    },
    "server": {
...
```
4. (Re-)Start KumaScript process (`node kumascript/run.js`)
    * [x] Should see no errors during startup
5. `tail -f newrelic_agent.log`
    * [x] After ~30s you should see connection logging messages show up

If this looks good, we will need to also:

* [ ] Update stage & prod `kumascript_settings_local.json` with the proper values for each environment.


jezdez:
Now that we're touching this code, I'd really like to adopt environment variables like we plan to do whenever we merge the settings refactor. nconf seems to support env vars as an alternative sources for those values (e.g. `newrelic:license_key` as the name of the env var). Ideally we'd also use a dotenv style thing but I'm not sure how easy it is to apply to nconf. Anyways, please take this into consideration when working on this.

lib/kumascript/conf.js
```diff
@@ -0,0 +1,66 @@
...
+            "https://developer.mozilla.org/en-US/docs/{path}?raw=1",
+        template_url_template:
+            "https://developer.mozilla.org/en-US/docs/en-US/Template:{name}?raw=1",
+        template_cache_control: 'max-age=3600'
+    },
+    repl: {
+        enabled: true,
+        host: "127.0.0.1",
+        port: 9070
+    }
```
jezdez: I think we should have sane defaults for newrelic here, so values for `app_name`, `high_security` and `license_key` -- for the latter an empty string or something that we can check for in `server.js`.

groovecoder: Good idea. It also means we don't need any corresponding update to `kumascript_settings_local.py` in the kuma repo.

groovecoder:
We spent a couple hours yesterday trying to use environment variables for these values and [we ran into issues](https://bugzilla.mozilla.org/show_bug.cgi?id=1005013#c12). So, I'm adopting the existing config methodology because:

1. I want to get New Relic monitoring on KumaScript now, so we can see into its operation
2. I want to move all the kumascript config into environment variables at one time, so we don't split-brain the config data

1 is more important than 2 right now. So, I filed [Bug 1201500](https://bugzilla.mozilla.org/show_bug.cgi?id=1201500) - Refactor settings into environment variables.

groovecoder:
Updated with default values for new relic in `lib/kumascript/conf.js`.

jezdez:
> On 03 Sep 2015, at 15:42, luke crouch <notifications@github.com> wrote:
> 
> We spent a couple hours yesterday trying to use environment variables for these values and we ran into issues. So, I'm adopting the existing config methodology because:
> 
> I want to get New Relic monitoring on KumaScript now, so we can see into its operation
> I want to move all the kumascript config into environment variables at one time, so we don't split-brain the config data
> 1 is more important than 2 right now. So, I filed Bug 1201500 - Refactor settings into environment variables.
> 
Cool, that makes sense to me. Thanks!

groovecoder:
Also updated description with local spot check instructions for @darkwing.

darkwing:
Instructions worked perfectly for me -- next step?

groovecoder:
Next step is in https://bugzilla.mozilla.org/show_bug.cgi?id=1005013 - we tried to push this to the stage celery node kumascript process and it fails to start for some reason we can't figure out. So we've got a support ticket opened with New Relic.

groovecoder:
We got stage ready for this, so I'm merging to prep a down-stream PR for Kuma to use this code.

PR #56, opened 2016-05-03
Add Dockerfile and Makefile

jgmize:
@jwhitlock r?

### DIFF
--- a/Dockerfile
+++ b/Dockerfile
@@ -1,0 +1,14 @@
+FROM debian:jessie
+
+ENV PYTHONDONTWRITEBYTECODE=1
+
+# need python and build-essential for node-gyp
+RUN apt-get update && \
+    apt-get install -y --no-install-recommends python2.7 build-essential nodejs npm
+RUN update-alternatives --install /usr/bin/node node /usr/bin/nodejs 10
+
+WORKDIR /app
+COPY . /app
+RUN npm install
+
+
--- a/Dockerfile
+++ b/Dockerfile
@@ -7,8 +7,9 @@ RUN apt-get update && \
     apt-get install -y --no-install-recommends python2.7 build-essential nodejs npm
 RUN update-alternatives --install /usr/bin/node node /usr/bin/nodejs 10
 
-WORKDIR /app
-COPY . /app
+ENV NODE_PATH=/node_modules
+COPY package.json /
 RUN npm install
 
-
+WORKDIR /app
+COPY . /app
--- a/Makefile
+++ b/Makefile
@@ -1,0 +1,44 @@
+VERSION ?= $(shell git describe --tags --exact-match 2>/dev/null || git rev-parse --short HEAD)
+REGISTRY ?= quay.io/
+IMAGE_PREFIX ?= mozmar
+IMAGE_NAME ?= kumascript
+IMAGE ?= ${REGISTRY}${IMAGE_PREFIX}/${IMAGE_NAME}\:${VERSION}
+MOUNT_DIR ?= $(shell pwd)
+APP_DIR ?= /app
+PORT ?= 9080
+DOCKER_RUN_ARGS ?= -v ${MOUNT_DIR}\:${APP_DIR} -w ${APP_DIR}
+DOCKER_PORT_ARGS ?= -p "${PORT}:${PORT}"
+DEIS_PROFILE ?= usw
+DEIS_APP ?= kumascript-dev
+PRIVATE_IMAGE ?= ${PRIVATE_REGISTRY}/${DEIS_APP}\:${VERSION}
+
+
+build:
+	docker build -t ${IMAGE} .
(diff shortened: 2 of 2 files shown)


jwhitlock:
What do you want out of a review?

jgmize:
> What do you want out of a review?

1. Install docker
2. `make build`
3. `make test`
4. `make run`
5. verify that running kumascript service works as expected on port 9080.

jwhitlock:
Failed on ``make build``: https://pastebin.mozilla.org/8870483

jgmize:
@jwhitlock that was just a network glitch: `Failed to fetch http://httpredir.debian.org/debian/pool/main/n/node-read/node-read_1.0.5-1_all.deb  Error reading from server. Remote end closed connection` -- would you mind trying again?

jwhitlock:
[Next ran passed](https://pastebin.mozilla.org/8870494). Moving on...

jwhitlock:
Failed on ``make test``: https://pastebin.mozilla.org/8870495

jgmize:
thanks @jwhitlock, it was passing locally b/c I ran `npm install` while the app dir was mounted against my local machine, which installed them into `node_modules` on the host machine, which were then being used during `make test`. I've added https://github.com/mozilla/kumascript/pull/56/commits/c146eae05858a168b14704a194cff78139d5952c so that it will use `/node_modules` within the container instead.

jwhitlock:
``make test`` works now.

In the development environment, you'd make a request like this to render the content of https://developer-local.allizom.org/en-US/docs/Mozilla:

```
curl  http://localhost:9080/docs/en-US/Mozilla
```

When running this against Docker, I get:

```
curl: (7) Failed to connect to localhost port 9080: Connection refused
```
There is no logging output from the ``make run`` command.

The other production difference is that we create a ``kumascript_settings_local.json`` file, with something like:

```
{
    "log": {
        "console": true,
        "file": {
            "filename": "/home/vagrant/logs/kumascript.log",
            "maxsize": 500000
        }
    },
    "server": {
        "port": 9080,
        "numWorkers": 4,
        "workerTimeout": 10000,
        "memcache": {
            "server": {"127.0.0.1:11211": 1},
            "options": { }
        },
        "macro_timeout": 60000,
        "document_url_template": "https://localhost/en-US/docs/{path}?raw=1",
        "template_url_template": "https://localhost/en-US/docs/en-US/Template:{name}?raw=1",
        "template_class": "EJSTemplate",
        "autorequire": {
            "mdn": "MDN:Common",
            "Date": "DekiScript:Date",
            "Page": "DekiScript:Page",
            "String": "DekiScript:String",
            "Uri": "DekiScript:Uri",
            "Web": "DekiScript:Web",
            "Wiki": "DekiScript:Wiki"
        }
    }
}
```
I don't know if any effort has been made to instead get config variables from the environment.  The code suggests port ``9000`` is the default in [one place](https://github.com/mozilla/kumascript/blob/master/lib/kumascript/server.js#L45), and ``9080`` in [another place](https://github.com/mozilla/kumascript/blob/master/lib/kumascript/conf.js#L30).

jgmize:
> curl  http://localhost:9080/docs/en-US/Mozilla
> When running this against Docker, I get:
> curl: (7) Failed to connect to localhost port 9080: Connection refused

The localhost port mapping works in linux and in the latest docker for mac beta (I recommend [signing up](https://beta.docker.com/) for it if you haven't already), but assuming you're using the current release of docker-toolbox then you'll need to connect to the virtualbox vm's ip on that port, which can be obtained with `docker-machine ip`. 

Locally for me, `make run` followed by `curl  http://localhost:9080/docs/en-US/Mozilla` in another shell gives me:
```
<p>The articles below include content about downloading and building Mozilla code. In addition, you'll find helpful articles about how the code works, how to build add-ons for Mozilla applications and the like.</p>

<p><p><strong>There are no subpages at this time.</strong></p></p>
```

> I don't know if any effort has been made to instead get config variables from the environment.

I haven't made any in this PR, but we can and should follow up on that after this PR lands.

jwhitlock:
``docker-machine ip default`` gives ``192.168.99.100``, and ``curl http://192.168.99.100/docs/en-US/Mozilla`` gives the same response.  It appears to be talking to production, since giving a recent page path ``curl http://192.168.99.100:9080/docs/id/Web/CSS`` gives the content of a [page created today](https://developer.mozilla.org/id/docs/Web/CSS).

+r, merge when ready.

groovecoder:
FWIW, I had [a commit on the heroku branch](https://github.com/mozilla/kumascript/commit/9279bfcb14953f579538141d28314f4f2bd188cb) that made kumascript listen to process.env.port.

Our [`nconf` module](https://www.npmjs.com/package/nconf) should be able to use environment variables for config values, but it seems like I had trouble getting the env vars to work?

PR #75, opened 2016-12-23
Bug 1325352 - Do nothing in WebExtCompat macro in preview mode.

arai-a:
slug is undefined in preview mode, so added the check in WebExtCompat macro to bailout.


SebastianZ:
It's good to fix the error. Though I think it would have been even better if there was a note about this like when using the {{cssinfo}} macro.

Sebastian

PR #99, opened 2017-01-26
bug 1332828: Avoid using DekiScript:Date

jwhitlock:
Switch some uses of the ``DekiScript:Date`` module to native ``Date()`` methods.


jwhitlock:
I tried removing the file, and ``page.subpages`` calls failed on a local copy of https://developer.mozilla.org/en-US/docs/MDN/Doc_status/JavaScript.  I am not sure why, but I decided to leave it for a bit.

PR #106, opened 2017-02-06
bug 1333120: fix headers in kumascript response

escattone:
**Background**
When we upgraded Node.js to 6.9.2  (#69) , we also upgraded to version 4 of [express](https://github.com/expressjs/express). Version 0.1.0 (the only version) of the [firelogger](https://github.com/lmorchard/node-firelogger) node module does not work with the latest API's of express' [Request](http://expressjs.com/en/4x/api.html#req) and [Response](http://expressjs.com/en/4x/api.html#res) objects, so it was causing problems (e.g., the "Vary" header was no longer correct).

**git comments**
* the firelogger npm module 0.1.0 does not work
  with version 4 of express (firelogger has not
  been updated in 5+ years)
* pull firelogger.js into the repo, update the
  request and response objects to use the version
  4 express API's, and fix for clean pass of jshint
* remove firelogger from the package.json and
  create a new npm-shrinkwrap.json
* fix broken test in test-server.js that should
  have detected this issue immediately


escattone:
FYI. I also submitted a [PR to the firelogger repo](https://github.com/lmorchard/node-firelogger/pull/1). If that is accepted and a new version of the module uploaded to npm, we can consider using that instead of the copy we pulled into our repo.

lib/kumascript/firelogger.js
```diff
@@ -0,0 +1,83 @@
+/*jshint node: true*/
```
escattone: FYI. This is a copy of https://github.com/lmorchard/node-firelogger/blob/master/firelogger.js but updated to use the latest methods of the Request and Response objects of `express` (specifically, the `get` and `set` methods). Also a comment has been added to the top of the file to suppress "implied global module" errors when running `jshint`.

lmorchard:
FWIW, there's no changelog on node-firelogger, because there was only ever one version :)

jwhitlock:
You're in good company. There were about 10 other node packages updated as well, and I stopped checking after 3 packages without changelogs.

PR #108, opened 2017-02-07
Fix bug 1335006

a2sheppy:
Change jsxref macro to use <%- %> for the displayed text, so that special
characters like "<" and ">" draw correctly instead of showing up as "&lt;"
etc.

### DIFF
--- a/macros/jsxref.ejs
+++ b/macros/jsxref.ejs
@@ -51,4 +51,13 @@ if (page && page.summary) {
 var code    = !$3 ? '<code>'  : '';
 var endcode = !$3 ? '</code>' : '';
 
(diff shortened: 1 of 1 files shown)


jwhitlock:
The text should be not be escaped if the ``<code>`` block is included (the default), and escaped if it is not included. I would expect:

```
{{jsxref("Operators/Comparison_Operators", "<", "#Less_than_operator")}}
{{jsxref("Ops", "<", "#Less", false)}}
```

to produce:
```
<a href="/en-US/docs/Web/JavaScript/Reference/Ops#Less" title="JavaScript has ..."><code>></code></a>
<a href="/en-US/docs/Web/JavaScript/Reference/Ops#Less" title="JavaScript has ...">&amp;lt;</a>

jwhitlock:
Manually merged the first commit to master.

PR #115, opened 2017-02-16
Add CSS4 Fonts to spec2.ejs

jpmedley:
For some reason this is already in SpecName.ejs.


SebastianZ:
Thanks Joe!

Sebastian

PR #116, opened 2017-02-19
Adding zh-CN and improvements

yfdyh000:
* Adding and improve translations for zh-CN locale, convert some `switch` to `mdn.localString`.

I cannot test it locally, due to there are some difficulties with "kuma" configure, like I cannot find the `kuma.sql.gz` file.

Todo: Convert more like `/switch.*(.+) {/`.


yfdyh000:
Dear @SphinxKnight, I'm not sure my practice is right or not, any of your views?

SphinxKnight:
Hi @yfdyh000 thanks :) I'll look into this in a few hours! (pinging @Elchi3 since this covers a lot of files)

macros/KeyRelatedEvents.ejs
```diff
@@ -1,18 +1,10 @@
...
-});
-
-var path2 = mdn.localString({
+var path = mdn.localString({
     "en-US" : "Mozilla_event_reference/",
     "ja"    : "Web/Reference/Events/"
 });
 
-var path = path1 + path2;
+var path = "/" + env.locale + "/docs/" + path2;
```
BychekRU: Please use `+ path` instead of `+ path2` or rename the variable above

macros/XULRefStyle.ejs
```diff
@@ -4,51 +4,22 @@
 
 var lang = env.locale;
```
BychekRU: You can safely remove this line

macros/draft.ejs
```diff
@@ -13,8 +13,34 @@
...
+    'ja': "このページは完成していません。",
+    'ko': "이 문서는 작성중입니다.",
+    'pl': "Strona ta nie jest jeszcze ukończona.",
+    'pt-BR': 'Esta página está incompleta.',
+    'pt-PT': 'Esta página está incompleta.',
+    'ru': 'Эта страница не завершена.',
+    'zh-CN': "本页尚未完工.",
+    'zh-TW': '本頁仍未完成'
+});
+var s_draft = ;
```
BychekRU: You define `s_draft`, `s_not_complete` above, delete these two lines

macros/csssyntax.ejs
```diff
@@ -64,23 +66,26 @@ var s_and = mdn.localString({
...
 var s_possible_values = mdn.localString({
     "en-US": "Possible values: ",
+    "zh-CN": "可能的值：",
     "de"   : "Mögliche Werte: ",
     "ru"   : "Возможные значения: "
 });
 
 var s_number_followed_by = mdn.localString({
-    "en-US": "a number followed by",
+    "en-US": "紧随数字",
```
BychekRU: You've translated string in en-US

macros/csssyntax.ejs
```diff
@@ -64,23 +66,26 @@ var s_and = mdn.localString({
...
     "de"   : "eine Nummer gefolgt von",
 });
 
 var s_like = mdn.localString({
     "en-US": "like",
+    "en-US": "像是",
```
BychekRU: Change locale code

yfdyh000:
@BychekRU Thank you for review this! updated.

yfdyh000:
@SphinxKnight @BychekRU

BychekRU:
@yfdyh000 I don't have locale installed MDN instance and reviewed your changes only by my eyes.
I'm strongly agree with your changes it really improves readibility.
Also I think it makes to use `mdn.localString` in `AddonSidebar.ejs` and `AppsLandingRightSide.ejs` as we do, for example, in `HTTPSidebar.ejs`. But we can do this later, or I can made it in my next PR.
By the way, I'm going to update Russian l10n in macros.
So this PR has a lot of changed files, and than faster we review and merge it than less conflicts it will has.
@Elchi3, please take a look

macros/A11yRoleQuicklinks.ejs
```diff
@@ -1,25 +1,6 @@
...
-        url = "/ja/docs/Web/Accessibility/AT-APIs/Gecko/Roles";
-        break;
-    case "ru":
-        url = "/ru/docs/Web/Accessibility/AT-APIs/Gecko/Roles";
-        break;
-    default:
-        url = "/en-US/docs/Web/Accessibility/AT-APIs/Gecko/Roles";
-        break;
-}
+var url = "/" + env.locale + "/docs/Web/Accessibility/AT-APIs/Gecko/Roles";
```
jwhitlock: This changes the behavior of 50+ other locales. I'm worried about the impact of that.

One option that does not change the behavior:

```
switch(lang) {
  case "de":
  case "fr":
  case "ja":
  case "ru":
  case "zh-CN":
    url = "/" + env.locale + "/docs/Web/Accessibility/AT-APIs/Gecko/Roles";
    break;
  default:
    url = "/en-US/docs/Web/Accessibility/AT-APIs/Gecko/Roles";
    break;
}
```

macros/CompatGeckoDesktop.ejs
```diff
@@ -5,6 +5,7 @@ var geckoVer = $0;
 
 var loca = mdn.localString({
     'en-US' : '1.7 or earlier',
+    'en-US' : '1.7 或更早',
```
jwhitlock: This should be ``zh-CN``, not ``en-US``

macros/FocusRelatedEvents.ejs
```diff
@@ -1,32 +1,10 @@
...
-
-%><ul>
-<li><a href="/en-US/docs/Web/Reference/Events/focus"><code>focus</code></a></li>
-<li><a href="/en-US/docs/Web/Reference/Events/blur"><code>blur</code></a></li>
-<li><a href="/en-US/docs/Web/Reference/Events/focusin"><code>focusin</code></a></li>
-<li><a href="/en-US/docs/Web/Reference/Events/focusout"><code>focusout</code></a></li>
-</ul><%
-
-} %>
+</ul>
```
jwhitlock: Add a newline to the end of this line

macros/L10n-SVG.json
```diff
@@ -11,6 +11,7 @@
     },
     "info_in_preview_not_available": {
         "en-US": "SVG info in preview not available",
+        "en-US": "预览模式下 SVG 信息不可用",
```
jwhitlock: This should be ``zh-CN``

macros/L10nErrors.ejs
```diff
@@ -1,23 +1,30 @@
...
-        inprogressStr = "Partiellement traduites";
-        woParentStr = "Non liées à une page en-US";
-    break;
-    default:
-        translatedStr = "Translated";
-        errorsStr = "Kumascript errors";
-        inprogressStr = "Partially translated";
-        woParentStr = "Not linked to an en-US page";
- }
+var str = mdn.localString({
```
jwhitlock: 👍 converting to localString. However, I haven't tried it to see if it works...

macros/LoadRelatedEvents.ejs
```diff
@@ -1,16 +1,5 @@
...
-switch (el) {
-    case "ja":
-    case "zh-TW":
-        path = "/" + el + "/docs/";
-        break;
-    default:
-        path = "/en-US/docs/"
-}
-
+var path = "/" + env.locale + "/docs/";
```
jwhitlock: Changes the default for all locales

macros/PlatformOnlyInline.ejs
```diff
@@ -15,26 +15,28 @@ if($1) {
     str = mdn.localString({
         "de": "Nur " + $0 + " "  + $1,
         "en-US": $0 + " "  + $1 + " Only",
-        "zh-CN": "只支持Gecko移动设备" + "(Gecko" + $1 + "以上)",
+        "zh-CN": "仅支持" + " "  + $1,
```
jwhitlock: Should there be a ``$0`` here? Otherwise, could be ``"仅支持 "  + $1``, no need to add strings to insert a space.

macros/A11yRoleQuicklinks.ejs
yfdyh000: Your practice changes its old behavior. I think it should apply to all locales, shows the localized user interface and page.

macros/LoadRelatedEvents.ejs
yfdyh000: This changes is expected. Any unexpected error should be added by a specific one, otherwise use this default.

yfdyh000:
Has rebase and squash.

@jwhitlock 
> I recommend closing this PR, and opening more, smaller PRs to cover the changes. Adding zh-CN cases to mdn.localString and case statements could be a fast merge, and several could appear in a single PR. Any algorithm changes should appear in their own PRs. However, if this is broken into 131 PRs (one for each file), that wouldn't be overkill.

I think a lot of PRs will be a nightmare for the review queue and commits messages.

If we find an error, there are any hindrances (such as caching) for revert a macro file and apply it to production? As well as any possibility to git pick this commit to the stage environment (https://developer.allizom.org/) as well as some testing / reviews?

macros/L10nErrors.ejs
yfdyh000: It is regrettable that seems no one to have a testing / review environment before merge commits.

jwhitlock:
After some discussion, the MDN team agrees that this change is too big. Please open smaller pull requests with only the string localizations, which will be fast to review and merge.

We are unlikely to merge pull requests that change the macro logic.

PR #146, opened 2017-03-31
Update legacy add-ons notice with concrete deprecation release info.

wbamberg:
Please see https://bugzilla.mozilla.org/show_bug.cgi?id=1340403. Since we've now published dates for when legacy add-ons will no longer be supported in Firefox, we should update the deprecation warning on MDN pages with this information.

### DIFF
--- a/macros/LegacyAddonsNotice.ejs
+++ b/macros/LegacyAddonsNotice.ejs
@@ -2,16 +2,17 @@
 
 var output = mdn.localString({
     'en-US': '<div class="warning">' +
-        '<p>We are planning to deprecate the use by Firefox add-ons of the techniques described in this document.</p>' +
-        '<p>Don\'t use these techniques to develop new add-ons. Use ' +
-        '<a href="/en-US/Add-ons/WebExtensions">WebExtensions</a> instead.' +
-        '<p>If you maintain an add-on which uses the techniques described here, consider migrating it to use ' +
-        'WebExtensions instead.<p>' +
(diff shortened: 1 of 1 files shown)


macros/LegacyAddonsNotice.ejs
```diff
@@ -2,16 +2,17 @@
...
-        '<p>Add-ons developed using these techniques might not work with multiprocess Firefox (e10s), which is ' +
-        'already the default in Firefox Nightly and Firefox Developer Edition, and will soon be the default in ' +
-        'Beta  and Release versions of Firefox. We have <a ' +
-        'href="/en-US/Add-ons/Working_with_multiprocess_Firefox">documentation on making your add-ons ' +
-        'multiprocess-compatible</a>, but it will be more future-proof for you to migrate to WebExtensions.</p>' +
+        '<p>Add-ons using the techniques described in this document are considered a legacy technology in Firefox. ' +
+        'Don\'t use these techniques to develop new add-ons. Use ' +
+        '<a href="/en-US/Add-ons/WebExtensions">WebExtensions</a> instead. ' +
+        'If you maintain an add-on which uses the techniques described here, consider migrating it to use ' +
+        'WebExtensions.<p>' +
```
jwhitlock: Should this be a closing tag ``</p>``?

wbamberg:
Thanks @jwhitlock ! I fixed that closing tag (plus another one), checked that the output was valid HTML, and re-tested the macro in Kuma.

PR #141, opened 2017-03-23
Fix FirefoxOSSidebar

jwhitlock:
Fix a bug in ``{{FirefoxOSSidebar}}`` where smart quotes in the German translation are causing rendering issues, such as [Archive/B2G_OS/Automated_testing/Endurance](https://developer.mozilla.org/en-US/docs/Archive/B2G_OS/Automated_testing/Endurance). Then format as valid JSON (double quotes) and indent.

PR #147, opened 2017-04-03
Fix FirefoxOSSidebar localizations (again)

jwhitlock:
Like #141, but actually tested locally to ensure it works.


jwhitlock:
Closing in favor of #152

PR #149, opened 2017-04-04
bug 1328989: Add /macros endpoint to get list of macros

jwhitlock:
``GET /macros/`` returns JSON with macro details, such as:

```
{
    "can_list_macros": true,
    "loader": "FileLoader",
    "macros": [
        {
            "filename": "A11yRoleQuicklinks.ejs",
            "name": "A11yRoleQuicklinks"
        },
        {
            "filename": "APIFeatureList.ejs",
            "name": "APIFeatureList"
        },
        ...
}
```
This can be used by Kuma to:

* Create a index of macros, similar to [/templates](https://developer.mozilla.org/en-US/docs/templates), when the templates are removed from the Kuma database
* Create links back to the macros in the KumaScript repo, for error messages and for viewing source

Adds a root endpoint too, because that bugged me.


jwhitlock:
I added the tests.

The ``test-server`` tests are strange to me - the ``fs.readFile`` appears necessary, or the test server will refuse the connection. The only explanation I could think of is that it needs 1/10 of a second to be ready for connections, but maybe there is something else going on.

Also, I'm sad that checklists are not checkable by anyone other than the author. 😢 I have repo access, so I could just edit the comment.

escattone:
Thanks for the tests @jwhitlock, they look great!

I was curious about your ``fs.readFile``/connection-refused comment. I imagine it arose when you were writing the test for the root URL as the expected output was small enough it didn't need to be read from a file? I couldn't reproduce the issue though. I removed the ``fs.readFile`` from the "Fetching the root returns the homepage" test and it worked (just setting ``expected`` directly as a string and going straight to the ``request(result_url, ...)`` call).

PR #150, opened 2017-04-05
Fixing a small French typo on Index.ejs macro

Porkepix:
There was a small typo on one of the French translations (wrong accent), this is fixing it.


wbamberg:
r+, works great, thanks for the contribution!

PR #159, opened 2017-04-14
bug 1354243: add macro linter (for JSON "macros")

escattone:
* expand the ``make lint-macros`` command to include the linting of JSON data files within the ``macros`` directory (using the newly added ``jsonlint-cli``)
* add ``make lint-macros`` to the "Test" stage of the Jenkins pipeline
* fix JSON syntax error in [macros/L10n-HTML.json](https://github.com/mozilla/kumascript/compare/add-json-linter-1354243?expand=1#diff-ca69a1de58c7bb296948e8113b837efa) so ``make lint-macros`` runs clean


escattone:
I wanted to also add a CSS linter (for the one CSS file within the ``macros`` directory - there are actually two files that have the ``.css`` extension, but ``macros/L10n-CSS.css`` is actually a JSON file and should probably use the ``.json`` extension instead?). I went down the path of trying ``stylelint``, but realized I was a fish-out-of-water when it came to setting up a strictly-syntax-error-only configuration. If we think thinks it's worthwhile, I can work later with @stephaniehobson to set that up.

stephaniehobson:
I did a quick look for a stylelint configuration we could borrow just to check for syntax errors but didn't turn anything up. I don't know how long it would take to put together a simple syntax check ourselves.

On the other hand, it could be argued that we should not have any CSS in this repo at all. Not even inline styles.

jwhitlock:
I agree:

* no CSS linting is required
* [L10n-CSS.css](https://github.com/mozilla/kumascript/blob/master/macros/L10n-CSS.css) should be renamed ``L10n-CSS.json``
* [CustomSampleCSS.css](https://github.com/mozilla/kumascript/blob/master/macros/CustomSampleCSS.css) should be deleted.  According to the [macros dashboard](https://developer.mozilla.org/en-US/dashboards/macros), it is unused on MDN.

escattone:
Thanks for the feedback @stephaniehobson  and @jwhitlock !

* renamed ``L10n-CSS.css`` to ``L10n-CSS.json``
* deleted ``CustomSampleCSS.css``

PR #160, opened 2017-04-17
Add two more APIs to GroupData.json

jpmedley:
* Credential Management
* Intersection Observer


macros/GroupData.json
```diff
@@ -950,6 +969,23 @@
             "properties": [],
             "events":     [ "message" ]
         },
+        "Intersection Observer API": {
+            "interfaces": [ "IntersectionObserver",
+                            "IntersectionObserverEntry" ],
+            "methods":    [ "IntersectionObserver.disconnect",
```
jwhitlock: Should these have parenthesis, like ``IntersectionObserver.disconnect()``? It seems the majority of methods do.

jpmedley:
PTAL

PR #162, opened 2017-04-25
cssinfo: Back to data/master, wrap initial values in <code>

jwhitlock:
In [properties.json](https://github.com/mdn/data/blob/master/css/properties.json), ``initial`` can have three types of values:

1. A literal value, like ``0`` or ``0% 0%`` or ``none``, as seen on [overflow](https://developer.mozilla.org/en-US/docs/Web/CSS/overflow).    
2. A key into the localization strings file, as seen on [-moz-appearance](https://developer.mozilla.org/en-US/docs/Web/CSS/-moz-appearance).
3. A list, each item which matches case 1 or 2.

Previously, case 1 (literal values) were designated with a ``<code>`` wrapper, and this was used in the MDN output. In https://github.com/mdn/data/pull/52, the wrapper was removed, so there is no longer a distinction between case 1 or 2.

This PR uses the presence of the string in the localization strings as a way to distinguish between cases 1 and 2.  If the string is not a localization key, a ``<code>`` wrapper is added for display.  The PR also normalizes tabs to spaces.

Alternatively, a stronger method could be used in properties.json to distinguish between case 1 and 2.

Related to PR #132.

### DIFF
--- a/macros/cssinfo.ejs
+++ b/macros/cssinfo.ejs
@@ -11,13 +11,13 @@ var name = $0 || (slug ? slug.split("/").pop().toLowerCase() :
     "preview-wiki-content");
 var atRule = $1;
 
-var propertiesUrl = "https://raw.githubusercontent.com/mdn/data/mdn/css/properties.json";
-var atRulesUrl = "https://raw.githubusercontent.com/mdn/data/mdn/css/at-rules.json";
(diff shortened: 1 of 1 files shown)


jwhitlock:
I think mdn/data#61 is going to take a while to resolve. I also suspect that we'll need to release a versioned NPM package sooner rather than later, since the mdn/data schema and data model continues to evolve, and KS doesn't currently have a clean way to upgrade data usage w/o breakage. Merging.

PR #164, opened 2017-04-26
LiveSampleURL: Make sure scheme is right on both prod and dev servers

a2sheppy:
Add code to the LiveSampleURL macro to make sure the scheme is right. In production, it should be https. Otherwise, it should match the one used for the main content since macros are loaded from the same domain as the wiki content in dev environments.

### DIFF
--- a/macros/LiveSampleURL.ejs
+++ b/macros/LiveSampleURL.ejs
@@ -8,7 +8,8 @@
 // $1   The URL from which to pull the sample; if not provided,
 //      the current page is assumed.
 
-var base = env.url;
+var pageUrl = env.url;
+var base = pageUrl;
 
 if ($1 && $1.length > 0) {
     base = $1;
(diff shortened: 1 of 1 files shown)


macros/LiveSampleURL.ejs
```diff
@@ -8,15 +8,29 @@
...
 
-// HACK: This may be the result of server misconfiguration. 
-base = base.replace('http://', 'https://'); 
+// HACK: This may be the result of server misconfiguration.
+// If we're on production, be sure we use https; otherwise,
+// use the same scheme as pageUrl.
+
+if (pageUrl.indexOf("developer.mozilla.org") != -1) {
+  // Production
+  base = base.replace('http://', 'https://');
```
jwhitlock: The scheme also needs to be ``https://`` on staging (developer.allizom.org)

jwhitlock:
Thanks @a2sheppy - it will be a lot easier to test live samples in a development environment now!

PR #165, opened 2017-04-27
Make csssyntax.ejs only link to named colors and system colors

wbamberg:
This is the first of several improvements I'd like to make to the "formal syntax" in CSS pages.

There are lots of color keywords, and currently every CSS property that can take a color value (which is also a lot) gets the same collection of keywords included in its formal syntax. This plays a big role in making the formal syntax a daunting wall of text: https://developer.mozilla.org/en-US/docs/Web/CSS/color#Formal_syntax, https://developer.mozilla.org/en-US/docs/Web/CSS/border#Formal_syntax.

Color keywords are syntactically very simple, so the formal syntax isn't explaining much actual syntax, and it's probably more helpful to link to a page that provides richer information (like a swatch and RGB values) than to just repeat the list on every page.

So that's what this change does: for "named_color" and "deprecated_system_color", it links to the relevant bit of https://developer.mozilla.org/en-US/docs/Web/CSS/color_value, and does not include the full list of values in the formal syntax.

So as not to look like so much of a terrible hack, it does this by adding a new data structure "externallyDescribedTypes": for types that appear in this structure, the macro should link to a resource external to the formal syntax, and not include details in the formal syntax itself.

### DIFF
--- a/macros/csssyntax.ejs
+++ b/macros/csssyntax.ejs
@@ -14,12 +14,32 @@ var syntaxesUrl = "https://raw.githubusercontent.com/mdn/data/master/css/syntaxe
 var unitsUrl = "https://raw.githubusercontent.com/mdn/data/master/css/units.json";
 // List types for which we want to link to a description instead of
 // including it in the formal syntax block
+var s_named_color_link = mdn.localString({
+    "en-US": "/docs/Web/CSS/color_value#Color_keywords",
+    "fr"   : "/docs/Web/CSS/Type_color#Les_mots-clés",
+    "de"   : "/docs/Web/CSS/Farben#Farbschlüsselwörter",
+    "es"   : "/docs/Web/CSS/color_value#Color_keywords",
+    "ja"   : "/docs/Web/CSS/color_value#Color_keywords",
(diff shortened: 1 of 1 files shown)


stephaniehobson:
+1 to removing the full list of colour names from the Formal Syntax

wbamberg:
There is a problem with l10n here. AIUI, if you just link to a page, like https://developer.mozilla.org/en-US/docs/Web/CSS/color_value, then l10n redirects are taken care of for you. But in this PR I wanted to link to a fragment: https://developer.mozilla.org/en-US/docs/Web/CSS/color_value#Color_keywords, in which case it seems they are not. So there's this clumsy and fragile attempt to handle different locales in this PR.

It would probably be OK to just link to https://developer.mozilla.org/en-US/docs/Web/CSS/color_value, then we could simplify this. Or we could split https://developer.mozilla.org/en-US/docs/Web/CSS/color_value#Color_keywords into its own page.

wbamberg:
Thanks @Elchi3 ! Colors are the big one, but there might be others. I'd really like to pretty-print formal syntax, but from my initial experiments that seemed a bit complicated.

PR #171, opened 2017-05-04
Macro ReleaseChannelInfo: use correct preposition for different date formats

a2sheppy:
Added logic so that if the release date parameter has two or more spaces in it,
it's treated as an exact date. When this happens, the preposition in the text
"will ship in <date>" was previously outputting "will ship in April 27, 2017".
The new logic fixes it to output "will ship on April 27, 2017".


wbamberg:
Why don't we throw away the date parameter and just link to https://wiki.mozilla.org/RapidRelease/Calendar, which is always updated when release dates change?

wbamberg:
As discussed, let's stop using this macro.

PR #190, opened 2017-05-30
Split all-examples from examples; es6-ify; change presentation for al…

wbamberg:
…l-examples

[webextensions-examples](https://github.com/mdn/webextensions-examples)  contains a file "examples.json" that contains information about the examples. The macro [WebExtExamples.ejs](https://github.com/mozilla/kumascript/blob/master/macros/WebExtExamples.ejs) does two quite distinct things with that data:

1) generate an overview page listing all the examples: https://developer.mozilla.org/en-US/Add-ons/WebExtensions/Examples

2) generate lists of links from WebExt API reference pages to the examples that illustrate those APIs, like this: [tabs.executeScript](https://developer.mozilla.org/en-US/Add-ons/WebExtensions/API/tabs/executeScript#Example_add-ons).

I've been thinking for a while that the presentation of the overview page is not ideal. On the [dev-addons list](https://mail.mozilla.org/pipermail/dev-addons/2017-May/002779.html), someone suggested using a tabular format for that page. I [tried it out](http://imgur.com/a/1bqo3?), and think it's a big improvement.

So that's the impetus to this PR. Initially I just updated the part of the old macro that built the overview page, using nice ES6 features to simplify it. But then I thought I should really ES6-ify the whole thing, rather than leave the macro inconsistent. But then I thought that since the macro is doing two quite different things with no real shared code, it would be better to split it. The also makes the code that figures out what to do less magical.

So now there's:
* a new macro "WebExtAllExamples.ejs" that just builds the overview page as a table
* modifies "WebExtExamples.ejs" to be ES6-y, and simplifies it by removing the overview-page stuff.

### DIFF
--- a/macros/WebExtAllExamples.ejs
+++ b/macros/WebExtAllExamples.ejs
@@ -1,0 +1,60 @@
+
+<%
+
+/*
+The "examples.json" file from the https://github.com/mdn/webextensions-examples
+repository contains information about the examples living in that repository.
+
+This macro loads the "examples.json" file and uses it to build a table
+presenting that information.
+*/
+
+const examplesBaseUrl = "https://github.com/mdn/webextensions-examples/tree/master/";
+const allExamples = mdn.fetchJSONResource("https://raw.githubusercontent.com/mdn/webextensions-examples/master/examples.json");
+const lang = env.locale;
+
+const s_name = mdn.localString({
+    "en-US": "Name"
+});
+
+const s_desc = mdn.localString({
+    "en-US": "Description"
+});
+
+const s_javascript_apis = mdn.localString({
+    "en-US": "JavaScript APIs"
+});
+
+function writeJavaScriptAPIs(apisJSON) {
+  let output = "";
+  for (let jsAPI of apisJSON) {
+    var link = `/${lang}/Add-ons/WebExtensions/API/${jsAPI.replace(".", "/")}`;
+    output += `<a href="${link}"><code>${jsAPI}</code></a><br/>`;
+  }
+  return output;
+}
+
+function writeAllExamples(examples) {
+  let output = "<table class=\"standard-table fullwidth-table\">";
+  output += `<tr>
+               <th>${s_name}</th>
+               <th>${s_desc}</th>
+               <th>${s_javascript_apis}</th>
+            </tr>`;
+
+  for (let example of examples) {
+    output += `<tr>`;
+    output += `  <td><a href="${examplesBaseUrl + example.name}">${example.name}</a></td>`;
+    output += `  <td>${example.description}</td>`;
+    output += `  <td>${writeJavaScriptAPIs(example.javascript_apis)}</td>`;
+    output += `</tr>`;
+  }
+  output += "</table>";
+  return output;
+}
+
+var output = writeAllExamples(allExamples);
+
+%>
+
+<%- output %>
--- a/macros/WebExtAllExamples.ejs
+++ b/macros/WebExtAllExamples.ejs
@@ -1,4 +1,3 @@
-
 <%
 
 /*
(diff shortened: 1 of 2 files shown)


jwhitlock:
I'll have bandwidth later today or early tomorrow to review this.

macros/WebExtAllExamples.ejs
```diff
@@ -0,0 +1,60 @@
...
+});
+
+const s_javascript_apis = mdn.localString({
+    "en-US": "JavaScript APIs"
+});
+
+function writeJavaScriptAPIs(apisJSON) {
+  let output = "";
+  for (let jsAPI of apisJSON) {
+    var link = `/${lang}/Add-ons/WebExtensions/API/${jsAPI.replace(".", "/")}`;
```
jwhitlock: Some of the locales (Korean, Greek) have translated the slug for Add-ons.  They are currently being standardized because of the vanity URL rewriting, but this will break if this feature is removed.  I think it is OK for today, we can let future devs figure it out.

macros/WebExtExamples.ejs
```diff
@@ -1,16 +1,47 @@
...
-var examplesInAPIHeading = $0 || "h3";
+This macro must be embedded in a WebExtensions API page, either of:
 
+* a "module" page (for example: "/en-US/Add-ons/WebExtensions/API/alarms" or "/en-US/Add-ons/WebExtensions/API/tabs")
+* an "API" page (for example: "/en-US/Add-ons/WebExtensions/API/alarms/create" or "/en-US/Add-ons/WebExtensions/API/tabs/executeScript")
+
+If it's embedded in a module page, it fetches all the examples that use any
+of the APIs in the given module, and outputs a list of links to those examples.
+
+If it's embedded in an API page, it fetches all the examples that use the
```
jwhitlock: This suggests to me that this macro could be further split into a version for module pages and a version for API pages, but I'm guessing that would require a lot more page editing, and wouldn't be worth it at this time.

macros/WebExtAllExamples.ejs
```diff
@@ -0,0 +1,60 @@
+
+<%
+
+/*
+The "examples.json" file from the https://github.com/mdn/webextensions-examples
+repository contains information about the examples living in that repository.
+
+This macro loads the "examples.json" file and uses it to build a table
+presenting that information.
+*/
```
jwhitlock: I like this intro comment, and would like to see this standard on macros.

Since this macro is basically for a single page, it would be good to include the URL https://developer.mozilla.org/en-US/Add-ons/WebExtensions/Examples in the comment.

macros/WebExtExamples.ejs
```diff
@@ -34,62 +74,35 @@ function getExamplesForAPI(module, api) {
...
-    output = "<" + examplesInAPIHeading + ">Example add-ons</" + examplesInAPIHeading + ">";
+    output = `<${desiredHeading}>${s_example_addons}</${desiredHeading}>`;
     output += "<ul>";
-    for (var i = 0; i < examples.length; i++) {
-      var example = examples[i];
-      output += "<li>";
-      var exampleURL = "https://github.com/mdn/webextensions-examples/tree/master/" + example.name;
-      output += '<a href="' + exampleURL + '">' + example.name + '</a>';
-      output += "</li>";
+    for (let example of examples) {
```
jwhitlock: This style of looping is a big improvement 👍

macros/WebExtAllExamples.ejs
```diff
@@ -0,0 +1,60 @@
+
```
jwhitlock: Unneeded space here.

jwhitlock:
@wbamberg I'll let you decide if the nits are worth fixing this round:

* Space at top of ``WebExtAllExamples``
* Add link to https://developer.mozilla.org/en-US/Add-ons/WebExtensions/Examples in the comment in ``WebExtAllExamples``
* ``mdn.localString`` treatment for the MDN API path in ``WebExtAllExamples``

jwhitlock:
Thanks for the updates! Let's get this on the train to production.

PR #206, opened 2017-06-13
bug 1372610: fix master Jenkins pipeline

escattone:
Fix the Jenkins pipeline for the master branch.

### DIFF
--- a/Jenkinsfile
+++ b/Jenkinsfile
@@ -8,7 +8,7 @@ node {
         }
 
         stage('Test') {
-          sh 'make lint-macros'
+          sh 'make lint-macros VERSION=latest'
           try {
             sh 'make test VERSION=latest TEST_RUN_ARGS="--reporter mocha-junit-reporter"'
           } finally {


jwhitlock:
That might work.  However, the [latest](https://quay.io/repository/mozmar/kumascript?tag=latest&tab=tags) in quay.io is also really out of date. Were you going to manually upload a new image?

It would also help to push the branch to the mozilla repo, so we don't have to guess if it will work 😄

escattone:
My understanding is that the ``make lint-macros VERSION=latest`` from the "Test" stage will use the local/cached Docker image built and tagged with ``quay.io/mozmar/kumascript\:latest`` from the "Build" stage (and not from quay.io), but maybe I'm off?

Since this bug only appears on the master branch, I didn't want to push to the mozilla repo, but I suppose I could put this branch in the ``Jenkinsfile`` (without the push step) and test it that way.

escattone:
I added a "fix-jenkins-master-pipeline-1372610" branch to the Jenkinsfile, pushed it to the mozilla repo for testing purposes (https://github.com/mozilla/kumascript/tree/fix-jenkins-master-pipeline-1372610), and it passed (see https://ci.us-west.moz.works/blue/organizations/jenkins/kumascript_multibranch_pipeline/detail/fix-jenkins-master-pipeline-1372610/1/pipeline).

PR #212, opened 2017-06-15
updating firefox_for_developers macro so that it doesnt break...

chrisdavidmills:
…for versions above Fx50.

Previously if you specified a number above 50 in the parameter, it would only show 50 as a maximum. I don't really see why it was set like this, so I've changed it.

### DIFF
--- a/macros/Firefox_for_developers.ejs
+++ b/macros/Firefox_for_developers.ejs
@@ -2,13 +2,13 @@
 <ul>
 <%
 /*
- * $0 max version to list
- *
+ * Parameters:
+ * $0 = max version to show in the list
  */
- 
-var CURRENT_MAX_VERSION = 50; 
- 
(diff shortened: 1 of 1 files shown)


jwhitlock:
It was changed back in Jan 2016, from a parameter to a fixed value:

https://github.com/mdn/archived_kumascript/commits/master/en-US/Firefox_for_developers.ejs
https://github.com/mdn/archived_kumascript/commit/ff59282fb323fb7e97c9e894c5de0e3707a1d2e3

I wonder if @teoli2003 recalls why this change was made, and maintained.

jwhitlock:
There are two pages that omit the parameter:

* https://developer.mozilla.org/en-US/docs/Mozilla/Firefox/Releases/2
* https://developer.mozilla.org/fr/docs/Mozilla/Firefox/Versions/22

Both appear to render, and display FF 3.6 and earlier, because JavaScript appears to treat an omitted parameter as 0 when doing math.  This seems like a bad idea, and the language should try harder in the future.

PR #214, opened 2017-06-15
bug 1372952: add liveness/readiness endpoints

escattone:
This PR adds "liveness" and "readiness" endpoints to KumaScript for use by Kubernetes or any similar control system.

* add "liveness"/"readiness" endpoints and their tests
* throw error if ``FileLoader`` is unable to find any macros during initialization, and add test for that case

I wrote the "readiness" endpoint under the idea that KumaScript is not ready unless it satisfies three things:
* it is up and running (the Express app can accept and respond to requests)
* it can load one or more macros
* it knows the document service (Kuma) is ready

Just for reference, Kuma's "readiness" only depends on two things:
* it is up and running (the Django app can accept and respond to requests)
* it can query the database

### DIFF
--- a/lib/kumascript/server.js
+++ b/lib/kumascript/server.js
@@ -2,7 +2,7 @@
 //
 // Provides the HTTP service for document processing
 
-/*jshint node: true, expr: false, boss: true */
+/*jshint node: true, esversion: 6, expr: false, boss: true */
 
 var ks_conf = require(__dirname + '/conf');
 
@@ -10,7 +10,9 @@ var newrelic_conf = ks_conf.nconf.get('newrelic');
 if (newrelic_conf && newrelic_conf.license_key) require('newrelic');
 
 // ### Prerequisites
-var _ = require('underscore'),
+var url = require('url'),
+    _ = require('underscore'),
+    request = require('request'),
     Memcached = require('memcached'),
     express = require('express'),
     morgan = require('morgan'),
@@ -126,6 +128,8 @@ var Server = ks_utils.Class({
             app.get('/docs/*', _.bind($this.docs_GET, $this));
             app.post('/docs/', _.bind($this.docs_POST, $this));
             app.get('/macros/?', _.bind($this.macros_list_GET, $this));
+            app.get('/healthz/?', _.bind($this.liveness_GET, $this));
+            app.get('/readiness/?', _.bind($this.readiness_GET, $this));
         });
     },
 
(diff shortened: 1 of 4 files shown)


tests/test-loaders.js
```diff
@@ -53,13 +54,31 @@ describe('test-loaders', function () {
...
+
     it('The FileLoader should detect duplicate macros', function () {
         assert.throws(
             function() {
                 new ks_loaders.FileLoader({
                     root_dir: 'tests/fixtures'
                 });
-            }
+            },
+            /duplicate macros:[\s\S]+/
```
jwhitlock: Regex assertion of error messages is interesting. Python doesn't have it, I think. This is the first thing I've been jealous of.

lib/kumascript/server.js
```diff
@@ -231,6 +235,52 @@ var Server = ks_utils.Class({
...
+    /**
+     * A "readiness" endpoint for use by Kubernetes or other
+     * similar systems. A successful response from this endpoint goes
+     * a step further and means not only that this Express app is up
+     * and running, but also that one or more macros have been found
+     * and that the document service is ready.
+     */
+    readiness_GET: function (req, res) {
+        var msg = 'service unavailable ',
+            parts = url.parse(this.options.document_url_template),
```
jwhitlock: I wonder if we should add this as a config item, or do similar work to parse ``document_url_template``, to fix [bug 1368134](https://bugzilla.mozilla.org/show_bug.cgi?id=1368134). No changes for this PR, just thinking of the next PR.

lib/kumascript/server.js
```diff
@@ -231,6 +235,52 @@ var Server = ks_utils.Class({
...
+            kumaReadiness = `${parts.protocol}//${parts.host}/readiness`;
+
+        // First, check that we can load some macros.
+        try {
+            // If there are no macros or duplicate macros, an error
+            // will be thrown.
+            this.macro_processor.makeLoader();
```
jwhitlock: 👍 works if I rename the ``macro`` folder

lib/kumascript/server.js
```diff
@@ -231,6 +235,52 @@ var Server = ks_utils.Class({
...
+        } catch(err) {
+            msg += `(macro loader error) (${err})`;
+            res.status(503).send(msg);
+            return;
+        }
+
+        // Finally, check that the document service is ready.
+        request.get(kumaReadiness, function (err, resp, body) {
+            if (!err && ((resp.statusCode >= 200) && (resp.statusCode < 400))) {
```
jwhitlock: 👍 works if I shut down the API server (``docker-compose stop api``)

lib/kumascript/server.js
```diff
@@ -231,6 +235,52 @@ var Server = ks_utils.Class({
...
+    /**
+     * A "liveness" endpoint for use by Kubernetes or other
+     * similar systems. A successful response from this endpoint
+     * simply proves that this Express app is up and running. It
+     * doesn't mean that its supporting services (like the macro
+     * loader and the document service) can be successfully used
+     * from this service.
+     */
+    liveness_GET: function (req, res) {
+        res.sendStatus(204);
```
jwhitlock: 👍 works

tests/test-server.js
```diff
@@ -21,6 +19,12 @@ describe('test-server', function () {
     beforeEach(function() {
         // Build both a kumascript instance and a document server for tests.
         this.test_server = ks_test_utils.createTestServer();
+        this.test_server.get('/healthz/?', function (req, res) {
```
jwhitlock: Does any code end up calling API's ``/healthz/``?

escattone: No, it doesn't, good catch!

lib/kumascript/server.js
escattone: Good question. I think I like the idea of a separate config item, as that might be clearer and more controllable/flexible?

escattone:
Thanks for the thorough review (as always) @jwhitlock!

PR #223, opened 2017-06-22
Create temporary folder for no macros test

jwhitlock:
Use the [tmp](https://www.npmjs.com/package/tmp) library to create and remove a temporary directory for the FileLoader test that detects no macros.  The previous method (creating a new folder inside the test folder) stopped working after TravisCI [updated their images](https://blog.travis-ci.com/2017-06-21-trusty-updates-2017-Q2-launch).


jwhitlock:
Well that didn't work either ☹️

escattone:
Thanks @jwhitlock for this attempt. Let's see what else might work.

jwhitlock:
We could pre-create an empty folder, but you can't check an empty folder into git - only files.  This is why you'll see ``.keep`` files or a readme.  However, the FileLoader loads all the files, including the placeholder, and will invalidate the test.  I tried having the FileLoader only load ``.ejs`` files, but then saw that we have ``.json`` and a ``.txt`` as well.  I regained my sanity before hard coding a whitelist of accepted extensions.

It appears that the FileLoader changes ``/tmp/tmp_XXXX`` to ``/app/tmp/tmp_XXXXX``.  Maybe we could make it not do that for root paths starting with a slash.

escattone:
Great idea @jwhitlock, that indeed paved the way for the rest of your ``tmp`` fix to do its thing. I submitted a PR to your branch that should give this PR the green light!

escattone:
I just noticed two things that need to change. I'll submit another PR to your branch. Coming soon...

escattone:
Ok, submitted https://github.com/jwhitlock/kumascript/pull/2 to your branch.

PR #235, opened 2017-07-12
Added German translation for TopicBox macro

SebastianZ:


macros/TopicBox.ejs
```diff
@@ -66,6 +66,12 @@ ircLink = "<a href='irc://irc.mozilla.org/" + ircChannelName + "'>#" + ircChanne
 // Generate localized strings
 
 switch(env.locale) {
+    case "de":  // German
+        headingStr = "Hilf dem '" + $0 + "'-Dokumentationsprojekt…";
+        statusStr = "Schau dir den aktuellen <a href='/en-US/docs/" + driver.docStatus + "'>Status</a> der '" + $0 +"'-Dokumentation an.";
```
jwhitlock: Should the URL be ``/de/docs/"?

macros/TopicBox.ejs
```diff
@@ -66,6 +66,12 @@ ircLink = "<a href='irc://irc.mozilla.org/" + ircChannelName + "'>#" + ircChanne
...
+        topicDriverStr = "Bereichsverantwortlicher: <a href='/en-US/profiles/" + driver.mdnPseudo + "'>" + driver.name + "</a> (IRC-Nickname: " + driver.ircPseudo + ")";
```
jwhitlock: Should the URL be '/de/profiles/'?

SebastianZ:
Sorry for the long delay on this one! I've changed the URLs of all localizations to the localized ones.

Sebastian

SebastianZ:
> I'm glad you figured out the rebase to make the PR commits sane.

Actually, I failed to do the rebase, because after rebasing I had plenty of unrelated commits, which came from the master branch. (I'd be happy if someone could explain me how that could happen.)

So I exported my changes into patches, deleted the local branch, recreated it, applied the patches, committed and force-pushed them.

Sebastian

jwhitlock:
@SebastianZ if you get into this situation again, get in touch with me. I can figure out what happened if you still have the commands you ran and their output, but I have no guesses why a rebase would go so badly.

OK, maybe one guess. ``git pull`` does the right thing 80% of the time, and messes things up the other 20%. You can avoid the 20% with some extra configuration:

https://stackoverflow.com/questions/15316601/in-what-cases-could-git-pull-be-harmful#15316602

PR #239, opened 2017-07-18
Add macro specific to interactive examples

stephaniehobson:
Creates iframe for interactive examples.

Also adds modifier classes based on:
- if it's a js example
- if modifier classes are passed when the macro is called

Classes are added to kuma in: https://github.com/mozilla/kuma/pull/4313 (but this is not dependent on that).

### DIFF
--- a/macros/EmbedInteractiveExample.ejs
+++ b/macros/EmbedInteractiveExample.ejs
@@ -1,0 +1,28 @@
+<%
+// Embeds a live sample from a mdn.github.io GitHub page
+//
+// Parameters:
+//  $0 - The URL of mdn.github.io page (relative)
+//  $1 - Keyword indicating iframe size: small, medium, or large
+//
+//  Example call {{EmbedInteractiveExample("interactive-examples/pages/css/animation.html", "small")}}
+//
+
+let url = "https://mdn.github.io/" + $0;
+let classesList = Array();
+
+// figure out classes
+if($0.includes('/js/')){
+    classesList.push('interactive-js');
+}
+
+if($1) {
+    classesList.push('interactive-' + $1);
+}
+
+let classes = classesList.join(' ');
+
+let str = `<iframe class="interactive ${classes}" width="100%" height="250" frameborder="0" src="${url}"></iframe>`;
+
+%>
+<%- str %>


macros/EmbedInteractiveExample.ejs
```diff
@@ -0,0 +1,31 @@
...
+if($1) {
+    classesList.push('interactive-' + $1);
+}
+
+let classes = '';
+for (var thisClass of classesList) {
+  classes += ' ' + thisClass;
+}
+
+let str = '<iframe class="interactive ' + classes + '" width="100%" height="250" frameborder="0" src="' + url + '"></iframe>';
```
wbamberg: Is the `height="250"` given here ignored and overridden by the `"interactive-[small, medium, large]"` class, if that argument is given?

How would this work for existing users of EmbedGHLiveSample, who pass a custom height? e.g. https://developer.mozilla.org/en-US/docs/Games (though I don't quite know why this uses this macro).

macros/EmbedInteractiveExample.ejs
```diff
@@ -0,0 +1,31 @@
...
+//
+//  Example call {{EmbedInteractiveExample("interactive-examples/pages/css/animation.html", "small")}}
+//
+
+let url = "https://mdn.github.io/" + $0;
+let classesList = Array();
+
+// figure out classes
+if($0.includes('/js/')){
+    classesList.push('interactive-js');
```
wbamberg: This indentation is inconsistent with line 25 (I don't know what the standard is, but I think I use 2 spaces).

macros/EmbedInteractiveExample.ejs
```diff
@@ -0,0 +1,31 @@
...
+}
+
+if($1) {
+    classesList.push('interactive-' + $1);
+}
+
+let classes = '';
+for (var thisClass of classesList) {
```
wbamberg: This could use `let` I guess.

macros/EmbedInteractiveExample.ejs
wbamberg: This might be clearer with a template literal.

macros/EmbedInteractiveExample.ejs
wbamberg: Could this also do classesList.join(' '); instead?

stephaniehobson: Ooooh, yes.

macros/EmbedInteractiveExample.ejs
stephaniehobson: Yes, the CSS will override the height defined in the HTML.

The custom height will be discarded. We can't do custom heights because there is no way to make the CSS change the custom height for mobile pages. 

I didn't know there were other users of `EmbedGHLiveSample` :/ Should I leave that macro alone or should that not be using the macro?

macros/EmbedInteractiveExample.ejs
stephaniehobson: Mozilla webdev standard is 4, I'll make sure I'm being consistant.

macros/EmbedInteractiveExample.ejs
wbamberg: > Should I leave that macro alone or should that not be using the macro?

Maybe it would be best to leave that macro.

stephaniehobson: Looks like there's more stuff than just the Interactive Examples using that macro, I will leave it alone.

macros/EmbedInteractiveExample.ejs
wbamberg: Thanks :). It's funny that this is different to the general standard for Mozilla code: https://developer.mozilla.org/en-US/docs/Mozilla/Developer_guide/Coding_Style#Indentation.

stephaniehobson: *face, palm*

Personally, I find 2 spaces as useful as 0. When scanning, the indent doesn't stand out to me at all. I wish tabs had won so we could all set our editors to show as much or as little indent as possible and didn't have to go from the tabs vs spaces argument to the 2 vs 4 spaces argument. But that ship has sailed :/

stephaniehobson:
@wbamberg Updated to more efficient code and to leave the live sample macro out of it.

PR #240, opened 2017-07-18
bug 1368134: add tests for parts of some auto-loaded macros

escattone:
This PR does two things:
- Improves the macro test framework such that it can test macros intended to be used as packages (e.g., [DekiScript-Page.ejs](https://github.com/escattone/kumascript/blob/master/macros/DekiScript-Page.ejs) and [DekiScript-Wiki.ejs](https://github.com/escattone/kumascript/blob/master/macros/DekiScript-Wiki.ejs)).
- Provides tests for the `buildAbsoluteURL()` and `uri()` functions within [DekiScript-Wiki.ejs](https://github.com/escattone/kumascript/blob/master/macros/DekiScript-Wiki.ejs) as well as the `subpages()`, `subpagesExpand()`, and `translations()` functions within [DekiScript-Page.ejs](https://github.com/escattone/kumascript/blob/master/macros/DekiScript-Page.ejs). This provides a baseline of regression tests for the refactoring of these functions within follow-on PR's (see #241).

### DIFF
--- a/tests/macros/test-dekiscript-page.js
+++ b/tests/macros/test-dekiscript-page.js
@@ -1,0 +1,310 @@
+/* jshint node: true, mocha: true, esversion: 6 */
+
+const fs = require('fs'),
+      path = require('path'),
+      sinon = require('sinon'),
+      utils = require('./utils'),
+      chai = require('chai'),
+      chaiAsPromised = require('chai-as-promised'),
+      assert = chai.assert,
+      itMacro = utils.itMacro,
+      describeMacro = utils.describeMacro,
+      beforeEachMacro = utils.beforeEachMacro,
+      fixture_dir = path.resolve(__dirname, 'fixtures');
+
+// Let's add "eventually" to assert so we can work with promises.
+chai.use(chaiAsPromised);
+
+// Load fixture data.
+const fixtures = {
+    spe: {
+        data: null,
+        filename: 'subpagesExpand-depth-gt-1.json'
+    },
+    spe0: {
+        data: null,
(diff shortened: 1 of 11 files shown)


tests/macros/utils.js
```diff
@@ -37,30 +41,74 @@ function createMacroTestObject(name, done) {
...
 
     /**
-     * Use this function to make test calls on the named macro. Its
-     * arguments become the arguments to the macro. It returns a promise.
+     * Wrap a package object such that when a property of that package (which
+     * we're assuming is a function for now) is requested, we'll wrap the
+     * function call in a Fiber. We have to run the function within a Fiber
+     * in order to mimic the enivironment in which it is normally called (
+     * macros are executed within a Fiber because it allows macro coders to
+     * write in a synchronous style).
```
jwhitlock: It sounds like getting rid of fibers will be harder than I assumed.

tests/macros/test-dekiscript-page.js
```diff
@@ -0,0 +1,310 @@
...
+    trans: {
+        data: null,
+        filename: 'translations.json'
+    }
+};
+for (const name in fixtures) {
+    fixtures[name].data = JSON.parse(
+        fs.readFileSync(
+            path.resolve(fixture_dir, fixtures[name].filename),
+            'utf8'
```
escattone: Good idea.

escattone:
Thanks @jwhitlock. I really want to get these fundamental "macros" fully tested and cleaned-up (e.g., I think the need for `DekiScript-Uri.ejs` can be eliminated), but that will have to wait until later.

PR #241, opened 2017-07-18
bug 1368134: fix page.subpagesExpand

escattone:
This PR depends on #240 to provide regression tests for the macro functions it refactors.

This PR fixes the problem unique to the Docker environment where macro functions like `subpagesExpand()` fail when making document-related requests because they use the host within the request URL (`env.url`) (e.g., `localhost:8000`) rather than the hostname of the proper container.
- extracts base URL for the "API" service (Kuma service dedicated to handling document requests from KumaScript) from the `document_url_template` setting.
- adds a `build_api_url()` function to each macro's context (globals) that builds an absolute URL to the "API" service.
- modifies the `subpages`, `subpagesExpand`, and `translations` functions within `DekiScript-Page.ejs` to construct URL's via `build_api_url()` rather than `env.url`.
- modifies the `buildAbsoluteURL` function within `DekiScript-Wiki.ejs`to construct URL's via `build_api_url()` rather than `env.url`.
- refactors the `uri` function within `DekiScript-Wiki.ejs` such that it continues to construct URL's using the request URL (`env.url`) since it's used to create links on pages.
- adds tests for the new `build_api_url()` function

### DIFF
--- a/lib/kumascript/server.js
+++ b/lib/kumascript/server.js
@@ -66,8 +66,12 @@ var Server = ks_utils.Class({
             app = this.app = express();
 
         this.req_cnt = 0;
-
         this.statsd = ks_utils.getStatsD(this.options);
+
+        // Build the root URL of the document API service.
+        this.doc_base_url = ks_utils.get_base_url(
+            this.options.document_url_template
+        );
 
         if (this.options.memcache) {
             var mo = this.options.memcache;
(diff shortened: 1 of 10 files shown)


lib/kumascript/server.js
```diff
@@ -256,8 +264,7 @@ var Server = ks_utils.Class({
      */
     readiness_GET: function (req, res) {
         var msg = 'service unavailable ',
-            parts = url.parse(this.options.document_url_template),
-            kumaReadiness = `${parts.protocol}//${parts.host}/readiness`;
+            kumaReadiness = url.resolve(this.doc_base_url, 'readiness');
```
jwhitlock: 😄 much cleaner now

escattone:
Thanks for the review @jwhitlock. I will rebase and re-run the tests. I think there will have to be some changes. Let me see.

escattone:
Submitted fix for `make test-macros`. I needed to set the `doc_base_url` option in `macro.ctx` for many of the new tests. All good to go!

PR #254, opened 2017-07-25
Add translations for pt-BR locale

Fernandolrs:
Translation, but a target function is required.


Fernandolrs:
So I observed, perhaps a Switch function to translate the next frame to the
Urls of destinations for each country.

But I noticed that URL conversion is automatic.

Thank you very much for your attention, I'll continue with the translations.

*Atenciosamente*

*Fernando Rubas*

2017-07-25 10:22 GMT-03:00 Florian Scholz <notifications@github.com>:

PR #261, opened 2017-07-27
Bug 1384891 - Add mdn.localStringMap and use it where localizing multiple texts.

arai-a:
https://bugzilla.mozilla.org/show_bug.cgi?id=1384891

### DIFF
--- a/macros/CanvasSidebar.ejs
+++ b/macros/CanvasSidebar.ejs
@@ -9,7 +9,7 @@ function state(section) {
   return "closed";
 }
 
-var text = mdn.localString({
+var text = mdn.localStringMap({
   'en-US': {
     'Canvas_API': 'Canvas API',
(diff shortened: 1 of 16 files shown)


arai-a:
here's the reason why I choose adding this function instead of defining all properties to other locales.

as discusses with @Elchi3 in IRC, it looks like fixing https://bugzilla.mozilla.org/show_bug.cgi?id=1378386 would take some more time,
and it's expected that people will keep adding more localized strings with current way (inside macro).

so both en-US text and localized text will be added incrementally meanwhile,
and it will become harder to keep them in sync, with "define all properties" way, especially with many properties case like HTTPSidebar,
and if they forgot to add corresponding text to localized map (object), it will result in issue like https://bugzilla.mozilla.org/show_bug.cgi?id=1384891 again.

instead, it would be nice that it becomes simply fallback text (en-US) even if the text is not localized.

PR #281, opened 2017-08-08
Background tasks is a candidate recommendation

stof:
https://www.w3.org/TR/requestidlecallback/


SebastianZ:
Thanks for the patch!

Sebastian

Issue #290, opened 2017-08-23
Add survey link and remove ability to pass a class


schalkneethling:
The EmbedInteractiveExample macro no longer needs to be able to accept a class so, this functionality can be removed.

We also need to add the copy, and link to the survey for beta testing the interactive editor.

stephaniehobson:
Fixed by https://github.com/mdn/kumascript/issues/290 👍

PR #291, opened 2017-08-23
Fix issue #290, remove class as optional argument and add survey link

schalkneethling:
Will open a PR for the styling related changes on Kuma.


stephaniehobson:
👍

PR #299, opened 2017-08-30
Update CSS Box Alignment.

jpmedley:


macros/SpecName.ejs
```diff
@@ -142,7 +142,7 @@ var specList = {
         url  : "https://drafts.csswg.org/css-box-3/"
     },
     'CSS3 Box Alignment':{
-        name : 'CSS Box Alignment Module',
+        name : 'CSS Box Alignment Module Level 3',
         url  : "https://drafts.csswg.org/css-align/"
```
SebastianZ: Can you please adjust the URL, too? It should rather be https://drafts.csswg.org/css-align-3/.

Sebastian

macros/spec2.ejs
```diff
@@ -40,7 +40,7 @@ var status = {
   'CSS3 Basic UI'              : 'CR',
   'CSS3 Basic UI 4'            : 'ED',
   'CSS3 Box'                   : 'WD',
-  'CSS3 Box Alignment'         : 'WD',
+  'CSS3 Box Alignment'         : 'ED',
```
SebastianZ: The status of the spec. is WD, see https://www.w3.org/TR/css-align-3/.

Sebastian

jpmedley:
PTAL

PR #301, opened 2017-08-31
bug 1395684: fix api-test for require()

escattone:
This fixes the api test "A template can import an npm module with require()", which started failing when the new `mdn-browser-compat-data` npm module, which has a new format, was released.
The test is using a test macro called `require-used` that is called with arguments assuming the old BCD format (e.g., the "version" and "data" keys), so I simply changed the arguments based on the new format.

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -6,6 +6,11 @@ services:
   - docker
 
 before_script:
+  # Remove the local, Travis-generated "node_modules" directory, because when
+  # we run the linters and tests, the local directory will be mounted into the
+  # container and its "node_modules" directory would override the one we want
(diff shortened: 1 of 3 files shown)


escattone:
I requested a review from @jwhitlock and @Elchi3, but only one is necessary.

escattone:
I removed the `node_modules` directory generated by Travis' `npm install` (which doesn't get an `npm update`) as that was overriding the `node_modules` directory we build at the root-level of the container (which does get an `npm update`), and allowed the Travis build to pass in earlier PR's when it should have failed, and fail in this PR when it should have passed.

PR #303, opened 2017-09-01
bug 1395745: add endpoint for serving revision hash

escattone:
This PR adds the `/revision` endpoint and associated test. It depends on https://github.com/mozilla/kuma/pull/4399. The endpoint returns the value of the environment variable `REVISION_HASH`, the git commit hash of HEAD baked into the kumascript Docker image at the time of the build.

### DIFF
--- a/tests/test-server.js
+++ b/tests/test-server.js
@@ -315,4 +315,12 @@ describe('test-server', function () {
             );
         });
     });
+
+    it('Revision endpoint returns git commit hash', function (done) {
+        testRequest(getURL('/revision'), done, function (resp, result) {
+            assert.equal(resp.statusCode, 200);
+            assert.equal(result, process.env.REVISION_HASH);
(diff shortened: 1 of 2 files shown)


escattone:
The Travis build will fail until mozilla/kuma#4399 is merged.

tests/test-server.js
```diff
@@ -315,4 +315,12 @@ describe('test-server', function () {
             );
         });
     });
+
+    it('Revision endpoint returns git commit hash', function (done) {
+        testRequest(getURL('/revision'), done, function (resp, result) {
+            assert.equal(resp.statusCode, 200);
+            assert.equal(result, process.env.REVISION_HASH);
```
jwhitlock: I get the string "undefined" back from the server, and the JS constant (?) ``undefined`` for ``process.env.REVISION_HASH``. If you force to a string, the test passes:

```
assert.equal(result, '' + process.env.REVISION_HASH);
```

Or, we could wait for PR 4399, and make sure it passes.

PR #306, opened 2017-09-06
bug 760240: provide interactive-examples settings to macros

escattone:
This PR:
- adds the interactive-example settings to kumascript, and makes them available in macros via the `env.interactive_examples` object
- changes the `EmbedInteractiveExample.ejs` macro to use the new setting, as well as provides tests for the macro
- is meant to replace the approach outlined in https://github.com/mozilla/kuma/pull/4404#issuecomment-327259243, which although [DRY](https://en.wikipedia.org/wiki/Don%27t_repeat_yourself) seemed like an unnecessarily complicated way to provide the setting to kumascript.
- does not provide the survey URL as part of the settings, since that seems only temporary per https://github.com/mozilla/kuma/pull/4404#issuecomment-327328890

### DIFF
--- a/lib/kumascript/server.js
+++ b/lib/kumascript/server.js
@@ -324,6 +324,11 @@ var Server = ks_utils.Class({
                 env.revalidate_at = (new Date()).getTime();
             }
 
+            // Add a clone of the interactive-examples settings to "env".
+            env.interactive_examples = ks_conf.nconf.get(
+                'interactive_examples'
+            );
+
             var ctx = {
                 env: env,
                 log: res.log
--- a/macros/EmbedInteractiveExample.ejs
+++ b/macros/EmbedInteractiveExample.ejs
@@ -10,7 +10,7 @@
 let jsEditorClass = "";
 let surveyClass = "interactive-editor-survey";
 let surveyLink = "https://www.surveygizmo.com/s3/3780959/MDN-Interactive-Editor";
-let url = "https://interactive-examples.mdn.mozilla.net/" + $0;
+let url = kuma.url.resolve(env.interactive_examples.base_url, $0);
 
 if ($0.includes('/js/')) {
     jsEditorClass = 'interactive-js';
--- a/lib/kumascript/conf.js
+++ b/lib/kumascript/conf.js
@@ -46,6 +46,9 @@ var DEFAULT_CONFIG = {
         enabled: true,
         host: "127.0.0.1",
         port: 9070
+    },
+    interactive_examples: {
+        base_url: "https://interactive-examples.mdn.mozilla.net"
     }
 };
 
--- a/tests/macros/test-EmbedInteractivExample.js
+++ b/tests/macros/test-EmbedInteractivExample.js
@@ -1,0 +1,41 @@
+/* jshint node: true, mocha: true, esversion: 6 */
+
+var utils = require('./utils'),
+    chai = require('chai'),
+    chaiAsPromised = require('chai-as-promised'),
+    assert = chai.assert,
+    itMacro = utils.itMacro,
+    describeMacro = utils.describeMacro;
+
+// Let's add "eventually" to assert so we can work with promises.
+chai.use(chaiAsPromised);
+
+describeMacro('EmbedInteractiveExample', function () {
+    itMacro('Typical settings and argument', function (macro) {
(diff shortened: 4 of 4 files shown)


escattone:
By the way, I'm sorry for the lack of camel-case. Python's practices are so embedded that I sometimes forget when I'm writing JavaScript. Feel free to suggest otherwise, and I'd be happy to change.

jwhitlock:
Yes, you can change the configuration by setting environment variables.  There's a [small section](https://kuma.readthedocs.io/en/latest/development.html#customizing-with-environment-variables) in the docs that mentions it.

With this change, I think the method would be:

1. Copy `.env-dist.dev` to `.env`, if you haven't already
2. Add a line:
  ```
  INTERACTIVE_EXAMPLES_BASE=http://localhost:8003
  ```
3. Restart the services with ``make up``

We'd need to update ``docker-compose.yml`` to set the container environment based on that variables.

In the worker section:

```
  - INTERACTIVE_EXAMPLES_BASE=${INTERACTIVE_EXAMPLES_BASE:-https://interactive-examples.mdn.mozilla.net}
```
In the kumascript section:

```
interactive_examples__base_url=${INTERACTIVE_EXAMPLES_BASE:-https://interactive-examples.mdn.mozilla.net}
```

If there's going to be a lot of development around these, then we may want to add an interactive examples container to ``docker-compose.yml``, or provide an alternate configuration.

escattone:
Thanks @jwhitlock !

jwhitlock:
I'm going to try adding a commit updating ``docker-compose.yml``, etc to https://github.com/mozilla/kuma/pull/4404

macros/EmbedInteractiveExample.ejs
```diff
@@ -10,7 +10,7 @@
 let jsEditorClass = "";
 let surveyClass = "interactive-editor-survey";
 let surveyLink = "https://www.surveygizmo.com/s3/3780959/MDN-Interactive-Editor";
-let url = "https://interactive-examples.mdn.mozilla.net/" + $0;
+let url = kuma.url.resolve(env.interactive_examples.base_url, $0);
```
jwhitlock: I'm getting an error:

```
Cannot read property 'base_url' of undefined
    at eval (eval at compile (/node_modules/ejs/lib/ejs.js:491:12), <anonymous>:21:52)
    at returnedFn (/node_modules/ejs/lib/ejs.js:520:17)
    at /app/lib/kumascript/templates.js:71:26
```

Adding some debug statements to see what is going on...

escattone: @jwhitlock Do you get this when you run `make test-macros`? Hmm, my first guess is that you may need to rebuild your kumascript image `cd .. && make build-kumascript`?

jwhitlock: No, the macros tests explictly set ``env.interactive_examples``, so the error doesn't occur. It only occurs when using the value set by ``server.js``

lib/kumascript/server.js
```diff
@@ -315,6 +315,11 @@ var Server = ks_utils.Class({
                 env.revalidate_at = (new Date()).getTime();
             }
 
+            // Add a clone of the interactive-examples settings to "env".
+            env.interactive_examples = _.clone(
```
jwhitlock: This should be:

```
env.interactive_examples = ks_conf.nconf.get('interactive_examples');
```

escattone: Oh man, good catch, sorry about that! Fix coming soon...

jwhitlock:
Also, I think [bug 760240](https://bugzilla.mozilla.org/show_bug.cgi?id=760240) is the wrong one. Maybe [bug 1308630](https://bugzilla.mozilla.org/show_bug.cgi?id=1308630)?

escattone:
@jwhitlock I amended my original commit to address your requested fix. Good catch, I should have tried it manually. 😞 I think it's good to go now?

escattone:
@jwhitlock Ugh, you're right, that bugzilla issue is not the right one, sorry! I'm not sure where I got that one from now that I think of it. Yes, I think [bug 1308630](https://bugzilla.mozilla.org/show_bug.cgi?id=1308630) will work fine, thanks. I'll amend the commit again to use that one.

PR #315, opened 2017-09-14
Update status of Payment Request API.

jpmedley:
It became a candidate recommendation today.

https://www.w3.org/blog/news/archives/6500


jpmedley:
Eeek! Don't know what I was doing there. Sorry.

PR #327, opened 2017-09-25
Fix spec reference for Navigator.share

jyasskin:
In https://developer.mozilla.org/en-US/docs/Web/API/Navigator/share


jpmedley:
Jeff,

They won't merge this unless it's also added to https://github.com/mdn/kumascript/blob/master/macros/spec2.ejs.

For this spec, I believe the status would be 'Living', but Mozilla folks will have to advise.

Joe

jpmedley:
And I just found your other PR. https://github.com/mdn/kumascript/pull/328

I usually do both in the same PR.

PR #328, opened 2017-09-25
Fix spec status for Navigator.share

jyasskin:
In https://developer.mozilla.org/en-US/docs/Web/API/Navigator/share


jpmedley:
Goes with: https://github.com/mdn/kumascript/pull/327

PR #382, opened 2017-10-31
Add Static Range spec.

jpmedley:


macros/SpecName.ejs
```diff
@@ -869,6 +869,10 @@ var specList = {
         name : 'SIMD',
         url  : 'https://tc39.github.io/ecmascript_simd/'
     },
+    'Static Range': {
+        name : 'Static Range',
+        url: : 'https://w3c.github.io/staticrange/index.html'
```
teoli2003: There is an extra : that prevent the tests to run successfully.

PR #384, opened 2017-11-01
Spanish translations (Added, fixed), litle refactoring

juan-ferrer-toribio:


juan-ferrer-toribio:
This request has code errors

PR #395, opened 2017-11-07
bug 1406546: add code for deploying kumascript to stage

escattone:
This PR requires https://github.com/mozilla/kuma/pull/4503. It adds the capability of automatic deployments of kumascript to stage when the `stage-push` branch is updated.

### DIFF
--- a/Jenkinsfile
+++ b/Jenkinsfile
@@ -1,21 +1,32 @@
+#!groovy
+
+@Library('github.com/mozmeao/jenkins-pipeline@master')
+
+utils = null
+
 node {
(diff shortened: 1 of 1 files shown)


Jenkinsfile
```diff
@@ -1,21 +1,32 @@
+#!groovy
+
+@Library('github.com/mozmar/jenkins-pipeline@master')
```
metadave: `github.com/mozmar` -> `github.com/mozmeao`

escattone: Good catch, I'll change that, thanks @metadave!

PR #402, opened 2017-11-10
Update HTMLSidebar macro for new HTML element page titles

a2sheppy:
For SEO reasons, the element pages' titles are being changed to the
form:

`<elementname>`: The Element Name element

This patch updates `HTMLSidebar` to strip off everything that isn't
the element name, so the title is what's expected by other parts of
the site. This lets the sidebar's list of HTML elements be correctly
formatted.

THIS PATCH REQUIRES THE `ListSubpagesForSidebar` PATCH (PR #401).


a2sheppy:
Just to have it on file, this PR no longer requires PR #401.

Issue #397, opened 2017-11-09
UI strings should be translated in Pontoon


jwhitlock:
*This is filed as [bug 1340342](https://bugzilla.mozilla.org/show_bug.cgi?id=1340342), and duplicated here to make it easier to link to in PRs, etc.*

Currently, user interface strings are translated inside macros, such as ``{{CompatNo}}``, which uses ``localString``:

https://github.com/mdn/kumascript/blob/f195712f5e9aba35268fd670def563c2f99c9f05/macros/CompatNo.ejs#L2-L12

It is difficult to update this translation or add new languages.  You need to open a PR, ensure the JSON is valid, etc. A responsible reviewer would ensure that the macro continues working in the new languages, and that it works in the new language. Translators tend to change many macros at once, which multiplies the effort to review the PR, and so translation PRs are often left open and unreviewed.

I'd like to make these strings available in Pontoon, so that instead ``CompatNo`` is:

```
var text=gettext('No&nbsp;support');
```

and the string ``No&nbsp;support`` would be available in [Pontoon](https://pontoon.mozilla.org/projects/mdn/) with the other MDN strings.

Some steps to get there:

* [x] Move translation files out of Kuma and into their own repository (https://github.com/mozilla-l10n/mdn-l10n)
* [x] Build KumaScript Docker image from the Kuma project so that locales will be available
* [ ] Write an EJS parser for [Babel](http://babel.pocoo.org/en/latest/messages.html), to extract strings from macros
* [ ] Integrate [Jed](http://messageformat.github.io/Jed/), [po2json](https://github.com/mikeedwards/po2json), and [sprintf-js](https://www.npmjs.com/package/sprintf-js) into KumaScript
* [ ] Convert some sample macros
* [ ] Document the process
* [ ] Recruit the community to convert remaining macros

PR #408, opened 2017-11-15
Adds fr l10n for event properties

Watilin:
Adds french localization for the macros about Event Properties.

I got some weird warning about line endings, but it doesn’t appear to has visible effect in the end. I’m sorry if I did something wrong.


jwhitlock:
@Watilin contacted me on IRC, and I talked about the issues around #397. I'll take the review on this one, since I approved creating it.

PR #412, opened 2017-11-20
Add Dutch translation

evelijn:


evelijn:
Sorry, I didn't read the previous pull requests before adding mine. This is probably not going to be pulled in due to the revamping, is that correct?

PR #431, opened 2017-11-29
wrap path argument to build_api_url with encodeURI

escattone:
Fixes: #307 
Replaces: #308, #422 

This PR corrects a regression I introduced in #241. I had thought that `url.resolve` would internally perform the same operation as `encodeURI`, not realizing that although it does encode characters like "<" and ">", it does not UTF-8 encode non-ASCII unicode characters like "é" as `encodeURI` does.

I also modified one of the existing tests for `build_api_url` to check for this case, so we can detect any future regressions.

This should fix many of the pages with errors: https://developer.mozilla.org/fr/docs/with-errors (and other non-English pages).

Thanks to @SphinxKnight and @Elchi3 (as well as apologies for being the cause of this issue! 😞 )

### DIFF
--- a/lib/kumascript/api.js
+++ b/lib/kumascript/api.js
@@ -387,7 +387,7 @@ var APIContext = ks_utils.Class({
         }
         return url.resolve(
             this.options.doc_base_url,
-            ks_utils.strip_base_url(path)
+            ks_utils.strip_base_url(encodeURI(path))
         );
     }
 });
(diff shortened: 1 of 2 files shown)


escattone:
Forgot to mention that I verified in my local dev environment that this PR does in fact fix #307 (using http://localhost:8000/fr/docs/Web/CSS/Valeur_calculée -- but I had to first scrape the documents into my local DB from production).

PR #433, opened 2017-12-01
bug 1422343: upgrade ejs to avoid security vulnerability

escattone:
Upgrade the Node.js package `ejs` from 2.5.2 to 2.5.7.


escattone:
Thanks @Elchi3 !

PR #443, opened 2017-12-11
Correct DOM3 Events link.

jpmedley:
'DOM3 Events' may be obsolete, but 'UI Events' has its own entry. Listing the later's location for the former is not a substitute for correcting MDN pages. Interfaces that are still part of a spec appear to only be part of an obsolete spec. The proper approach is to have the correct link for both and (if appropriate) link to both in MDN pages ([example here](https://developer.mozilla.org/en-US/docs/Web/API/KeyboardEvent/KeyboardEvent)). 

I'm updating the pages and am happy to let this PR sit until they are done.

### DIFF
--- a/macros/SpecName.ejs
+++ b/macros/SpecName.ejs
@@ -495,7 +495,7 @@ var specList = {
     },
     'DOM3 Events':{
         name : 'Document Object Model (DOM) Level 3 Events Specification',
-        url  : 'https://www.w3.org/TR/2009/WD-DOM-Level-3-Events-20090908/'
+        url  : 'https://www.w3.org/TR/2014/WD-DOM-Level-3-Events-20140925/'
     },
     'DOM3 XPath':{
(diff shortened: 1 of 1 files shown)


macros/SpecName.ejs
```diff
@@ -495,7 +495,7 @@ var specList = {
     },
     'DOM3 Events':{
         name : 'Document Object Model (DOM) Level 3 Events Specification',
-        url  : 'https://w3c.github.io/uievents/'
+        url  : 'https://www.w3.org/TR/2009/WD-DOM-Level-3-Events-20090908/'
```
myakura: there's a newer version at https://www.w3.org/TR/2014/WD-DOM-Level-3-Events-20140925/ (the last public WD named as DOM3 Events). is thre any reason pointing to the 2009 snapshot?

jpmedley: I thought they all redirected to UI Events. I was trying to eliminate confusion. Fix it now.

PR #468, opened 2017-12-15
remove unused macro toc

escattone:
Remove unused `toc` macro.


wbamberg:
To look for internal calls, how do you search for macros with common names like this? Does this work:

`grep -r \([\"\']toc ./macros`

?

escattone:
I always use the built-in grep of my atom editor, but I think you're `grep -r \([\"\']toc ./macros` is good, although perhaps it would be safer to strip off the leading `\(`, just so you don't have to handle possible spaces after, like `template( "toc")`.

PR #516, opened 2017-12-17
LearnSidebar German Translation Update

Shidigital:
Update to English Menu, 
from "Du" to "Sie" change, 
spelling misstakes, etc.


Shidigital:
Hi!
Thanks for the review.

I changed it and several articles to "Sie" after I asked in the IRC chat what they prefer me to do about it.
There I was told to use "Sie", otherwise I would have used "Du" as well. 

You are right, maybe best way would be to take it out where possible, even if this means that it differs from the English version.

Elchi3:
Thanks for asking around in IRC for more opinions. I'm merging this, but we can always improve to a more neutral language in the future.

PR #535, opened 2017-12-19
Add make test-macros to docs

jwhitlock:


escattone:
Thanks @jwhitlock and @Elchi3 !

PR #552, opened 2018-01-10
Adding the "Request the right permissions" page to the how-to section.

rebloor:
For more details, see https://trello.com/c/tcxh8b2W


wbamberg:
This seems like a very Firefox-specific article, perhaps it would be better in a different section: either Firefox workflow or Publishing? It feels to me like it might fit better into Publishing, as it's somehow about the user's installation experience.

rebloor:
You might have a point there Will, I’m checking with Mike.

Regards
Richard Bloor

PR #555, opened 2018-01-11
Fix issue #408 on interactive examples, allow optional class to be passed from macro

schalkneethling:
For context, please see https://github.com/mdn/interactive-examples/issues/408

### DIFF
--- a/macros/EmbedInteractiveExample.ejs
+++ b/macros/EmbedInteractiveExample.ejs
@@ -3,10 +3,12 @@
 //
 // Parameters:
 //  $0 - The URL of interactive-examples.mdn.mozilla.net page (relative)
+//  $1 - Optional custom height class to set on iframe element
 //
-//  Example call {{EmbedInteractiveExample("pages/css/animation.html")}}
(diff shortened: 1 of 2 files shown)


macros/EmbedInteractiveExample.ejs
```diff
@@ -3,18 +3,24 @@
...
 if ($0.includes('/js/')) {
     jsEditorClass = 'interactive-js';
 }
 
-let str = `<iframe class="interactive ${jsEditorClass}" width="100%" height="250" frameborder="0" src="${url}"></iframe>`;
+if ($1) {
+    customHeightClass = $1;
+}
+
+let str = `<iframe class="interactive ${jsEditorClass} ${customHeightClass}" width="100%" height="250" frameborder="0" src="${url}" title="MDN Web Docs Interactive Example"></iframe>`;
```
jwhitlock: This adds an extra space to the ``<iframe>`` that is causing tests to fail. Options are:

1. Add the extra space to the expected strings in the tests
2. Modify ``customHeightClass`` to include a leading space, and trim the space from ``let str``.

Sorry that we don't have an HTML normalizer for the tests 😢

PR #572, opened 2018-01-31
Remove non-existent "muted" event

a2sheppy:
The Media Capture and Streams API entry in GroupData.json
incorrectly had a "muted" event. This removes that.


a2sheppy:
Just to be explicit in saying so: I have indeed tested this locally both by linting and by importing appropriate content from MDN into my local Kuma and checking that the sidebar no longer includes "muted" but still includes the rest.

PR #574, opened 2018-02-02
Add WebAuthn spec

apowers313:
Added entry for [WebAuthn](https://w3c.github.io/webauthn/) to `SpecData.json`


apowers313:
I'm hoping this PR adds `WebAuthn` to the `specList` so that I can use `{{SpecName}}` to reference `WebAuthn` on [MDN](https://developer.mozilla.org/en-US/docs/Web/API/Credential_Management_API). Let me know if there's a better way to do this.

PR #582, opened 2018-02-05
Rm suivant

SphinxKnight:
This macro was a translation of Next.ejs. It was only used by fr pages where I changed those to Next calls.
I've browsed the codebase and it doesn't seem to appear. (at least not by the simple name "Suivant").
I think we can safely get rid of this one.

Previous commits were used for "refreshing" my fork and can be squashed.


jwhitlock:
I've searched the docs, and I can't find any active documents using this macro either.

PR #590, opened 2018-02-08
Update Japanese translation

mfuji09:


mfuji09:
Japanese translation seems older, so I updated the translation from the English message.

PR #611, opened 2018-02-14
Adding link to new "What next?" article in the "Getting started" section

rebloor:
For more information, see  https://trello.com/c/onTACvVX


rebloor:
Hi Will, an example of the dangers of assuming that what's there is a good template and not going back to first principles :-( I've fixed this one and have, hopefully, not gone too far in fixing the other lines that were missing closing tags.

PR #637, opened 2018-02-21
added chinese simplified translation for learnsidebar

allan-simon:


allan-simon:
sorry to ask but do you have an idea on when it will be available on MDN ? I'm teaching html/css to some chinese friends who don't speak a word of english and that would help them considerably to have the menu on the left in chinese :)  (currently they can't navigate and they have bookmarked all the pages to have the title in chinese )

SphinxKnight:
Hi @allan-simon, the change is now available on MDN. I've force-refreshed https://developer.mozilla.org/zh-CN/docs/learn/HTML and got the right sidebar.
The only issue is that there is no automatic "refresh" of the different pages so one need to log in and force refresh (ctrl-shift-r) the pages that  weren't updated yet. (force refresh doesn't work if not logged in)

SphinxKnight:
@allan-simon, thanks to @jwhitlock, the pages were re-rendered so that manual action isn't required anymore :) (cf. [this bug for more details](https://bugzil.la/1365987))

allan-simon:
@SphinxKnight  awesome, thanks @jwhitlock  :)

PR #639, opened 2018-02-28
trying to fix the guide links in the web components sidebar - third try!

chrisdavidmills:
I think these should actually work now. Can you check it for me @Elchi3  ?


chrisdavidmills:
Yowza, what a song ;-)

PR #640, opened 2018-03-02
bug 1413523 Switch to node.js 8.10.0

MatonAnthony:
The current version of node.js we use is limiting or abilities
to use the new features available.

The node project has released a second LTS cycle 8.9.4 which will
be supported until December 2019.

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -2,7 +2,7 @@ sudo: required
 
 language: node_js
 node_js:
-    - "6"
+    - "8"
 
 services:
   - docker
--- a/README.md
+++ b/README.md
@@ -140,7 +140,7 @@ more before you run your local development version of MDN.
 
 ## Setup (Standalone)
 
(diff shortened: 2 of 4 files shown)


jwhitlock:
When I build an image using this, I get a very different ``npm-shrinkwrap.json``. What was your process?

MatonAnthony:
I used the instruction in README.md.

jwhitlock:
The current node 6, or with node 8? I'm using node 6, current master.

MatonAnthony:
Since it's a `npm-shrinkwrap.json` for node 8, I generated after updating the image to node 8

jwhitlock:
OK, that makes sense. It works with node 6, so I believe this should be merged before https://github.com/mozilla/kuma/pull/4692

MatonAnthony:
Yes definitely. I'm investigating the Travis CI error.

PR #671, opened 2018-04-06
Update csssyntax.ejs

fALKENdk:
Check if syntax is undefined.

Macro on https://developer.mozilla.org/en-US/docs/Web/CSS/:matches does not work because syntax seems to be undefined.

```
Problem executing template csssyntax:
   35 | <h2 id="Syntax">Syntax</h2>
   36 | 
   37 | <pre class="syntaxbox">{{csssyntax}}</pre>
-------------------------------^
   38 | 
   39 | <h2 id="Examples">Examples</h2>

TypeError: ejs:1
 >> 1| <%
    2| /*
    3|   Displays the syntax of a CSS property or descriptor
    4| 

Cannot read property 'syntax' of undefined
    at eval (eval at compile (/node_modules/ejs/lib/ejs.js:549:12), <anonymous>:454:42)
    at returnedFn (/node_modules/ejs/lib/ejs.js:580:17)
    at /app/lib/kumascript/templates.js:71:26
```

### DIFF
--- a/macros/csssyntax.ejs
+++ b/macros/csssyntax.ejs
@@ -391,13 +391,15 @@ function formatTypesSyntax(formattedSyntax, describedTypes) {
                 // described before
                 // check whether https://github.com/mdn/data/pull/66 was merged
                 var syntax = data.syntaxes[type].syntax || data.syntaxes[type];
-                if (describedTypes.indexOf(type) === -1) {
-                    typesSyntax += "<span id=\"" + type + "\">&lt;" + type +
-                        "&gt;</span> = " + formatSyntax(syntax);
-                    if (index < types.length - 1) {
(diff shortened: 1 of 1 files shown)


fALKENdk:
Same with:
https://developer.mozilla.org/en-US/docs/Web/CSS/::-moz-color-swatch
https://developer.mozilla.org/en-US/docs/Web/CSS/:host
https://developer.mozilla.org/en-US/docs/Web/CSS/:host-context()
https://developer.mozilla.org/en-US/docs/Web/CSS/::slotted

fALKENdk:
You are absolutely right @wbamberg - I failed to test it before, sorry.
And then I deleted my fork ... so I had to make a new pull request here: #687

PR #681, opened 2018-04-07
Correct Japanese translation in spec2.ejs

mfuji09:
廃止された is unnatural as a status. 廃止 is more preferred.


SphinxKnight (approved):
Thanks @mfuji09! :)

PR #685, opened 2018-04-07
pt-BR translations for HTTPSidebar macro

GPrimola:


SphinxKnight (changes requested):
The locale identifier is not correct here (en-US instead of pt-BR I assume).
Could you fix this one please?

Apart from this, I cannot see any errors

SphinxKnight (approved):

PR #687, opened 2018-04-09
Check if selector exists in selectors.json

fALKENdk:
History in: #671


wbamberg:
@fALKENdk , so sorry to be so slow reviewing this. I'll get to it today.

PR #690, opened 2018-04-12
Provide tests for svginfo

JeremiePat:
Here are some test for the current state of {{svginfo}}

They cover:
- [x] calling svginfo in preview mode
- [x] calling svginfo for non existent SVG element
- [x] calling svginfo with an explicit parameter
- [x] calling svginfo with an implicit parameter (based on the page slug)
- [x] checking the various type of linking for categories and permitted contents

> **Note:** _This should be reviewed and merged before PR #670 that introduce changes to svginfo_

### DIFF
--- a/tests/macros/test-svginfo.js
+++ b/tests/macros/test-svginfo.js
@@ -1,0 +1,496 @@
+/* jshint node: true, mocha: true, esversion: 6 */
+
+// Get necessary modules
+const sinon = require('sinon');
+const { itMacro, describeMacro, beforeEachMacro } = require('./utils');
+const chai = require('chai');
+const chaiAsPromised = require('chai-as-promised');
+
+// Set up Chai
+chai.use(chaiAsPromised);
+
+// Basic const
+const SVG_BASE_SLUG = 'docs/Web/SVG';
+const SVG_DATA      = require('../../macros/SVGData.json');
+const L10N_COMMON   = require('../../macros/L10n-Common.json');
+const L10N_SVG      = require('../../macros/L10n-SVG.json');
+
+
+// UTILITIES
+// ----------------------------------------------------------------------------
+// NOTE: Maybe it would worth having those in a separate module
+//       if we want to deal with the same stuff in other tests.
+// ----------------------------------------------------------------------------
+
+// Handle l10n mechanics
+//
+// @param key    The key of the string to localize
+// @param locale The locale into which the key must be translated
+// @return <string>
+function _ (key, locale) {
+    if (typeof key === 'object' && key !== null && !Array.isArray(key))
+        return key[locale] || '';
+
+    key = String(key);
+
+    if (key in L10N_SVG)
+        return L10N_SVG[key][locale] || L10N_SVG[key]['en-US'] || '';
+
+    if (key in L10N_COMMON)
+        return L10N_COMMON[key][locale] || L10N_COMMON[key]['en-US'] || '';
+
+    // No matter what that macro must return a string
+    return '';
+}
+
+// Build an absolute URL by concatenating the arguments.
+function URL(...chunks) {
+    return '/' + chunks.join('/');
+}
+
+// Turn a camelCase string into a snake_case string
+//
+// @param str     <string>  The string to transform
+// @param upFirst <boolean> Indicate is the first letter must be upper cased (true by default)
+// @return <string>
+function camelToSnake (str, upFirst=true) {
+    str = str.replace(/[A-Z]/g, (match) => "_" + match.toLowerCase());
+
+    if (upFirst)
+        str = str.replace(/^./, (match) => match.toUpperCase());
+
+    return str;
+}
+
+// Extract the summary of a given page in a given locale if it exists
+//
+// @param key    <string>  The key of the page to retrieve
+// @param locale <string>  The locale in which the summary must be provided
+// @param clean  <boolean> Indicate if HTML tags must be striped (true by default)
+// @return <string>
+function getSummary(key, locale, clean=true) {
+    if (!MOCK_PAGES[locale] ||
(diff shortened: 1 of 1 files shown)


tests/macros/test-svginfo.js
```diff
@@ -0,0 +1,496 @@
...
+// ----------------------------------------------------------------------------
+// Each test case is define by:
+// A `title` to make the test understandable by a human behing
+// An `input` which is an Array of parameters that will be passed to the macro
+// An `output` which is the string that the macro should return,
+// A `env` that is overriding the default env variable inside the macro
+//
+// NOTE: we could probably make that more generic by having a single test
+//       runner (see below) and a bunch of JSON files (one per macro) to
+//       describe all the possible inputs and their expected outputs.
```
escattone: On further reflection, in order for this approach to be used for more general testing, the definition of each test case would also have to include its `beforeEachMacro` set-up (creating and populating the stubs for example), which seems difficult to standardize.

tests/macros/test-svginfo.js
```diff
@@ -0,0 +1,496 @@
...
+        Object.keys(MOCK_PAGES['en-US']).forEach((key) => {
+            const { url, data } = MOCK_PAGES['en-US'][key];
+            macro.ctx.wiki.getPage.withArgs(url).returns(data);
+        });
+
+        Object.keys(MOCK_PAGES['zh-CN']).forEach((key) => {
+            const { url, data } = MOCK_PAGES['zh-CN'][key];
+            macro.ctx.wiki.getPage.withArgs(url).returns(data);
+        });
+    });
```
escattone: This is nicely done, but it's missing the stub for the `wiki.pageExists` call that is made within the `SVGElement` macro (via the call to `template("SVGElement"...` made in one of the paths through the `svginfo` macro). The way the test runs currently, it's making an actual network request via the real `wiki.pageExists`. Something like this worked for me during my testing:
```javascript
    beforeEachMacro((macro) => {
        // let's make sure we have clean calls to wiki.getPage
        macro.ctx.wiki.getPage = sinon.stub();
        macro.ctx.wiki.pageExists = sinon.stub();

        Object.keys(MOCK_PAGES['en-US']).forEach((key) => {
            const { url, data } = MOCK_PAGES['en-US'][key];
            macro.ctx.wiki.getPage.withArgs(url).returns(data);
            macro.ctx.wiki.pageExists.withArgs(url).returns(true);
        });

        Object.keys(MOCK_PAGES['zh-CN']).forEach((key) => {
            const { url, data } = MOCK_PAGES['zh-CN'][key];
            macro.ctx.wiki.getPage.withArgs(url).returns(data);
            macro.ctx.wiki.pageExists.withArgs(url).returns(true);
        });
    });
```

This reminds me that I need to submit a PR that stubs the `request` package within the `macro.ctx` object, so that it throws an error when called. I think that would be the best way to catch all of the attempts to make real network calls, and help us discover what we need to stub.

tests/macros/test-svginfo.js
```diff
@@ -0,0 +1,496 @@
...
+
+    // Set the list of permitted content that apply to the element
+    const permittedContent = [DESCRIPTION];
+
+    if (data.content.elements) {
+        // Regroup permitted content between named groups and standalone elements
+        let { elements, groups } = data.content.elements.reduce((acc, value) => {
+            if (value.indexOf('&lt;') !== -1) {
+                let key     = value.replace(/&lt;|&gt;/g, '');
+                let url     = URL(locale, SVG_BASE_SLUG, 'Element', key);
```
escattone: Whitespace again: I would prefer only a single space before each `=`.

tests/macros/test-svginfo.js
```diff
@@ -0,0 +1,496 @@
...
+                let summary = getSummary(key, locale);
+
+                acc.elements.push(`<a href="${url}"  title="${summary}"><code>${value}</code></a>`);
+            } else {
+                let anchor = "#" + camelToSnake(value);
+                let url    = URL(locale, SVG_BASE_SLUG, 'Element') + anchor;
+                let label  = _(value, locale);
```
escattone: Whitespace again: I would prefer only a single space before each `=`.

tests/macros/test-svginfo.js
```diff
@@ -0,0 +1,496 @@
...
+
+                acc.groups.push(`<a href="${url}">${label}</a>`);
+            }
+
+            return acc;
+        }, {elements: [], groups: []});
+
+        // Named groups must be listed first (each on new lines)
+        if (groups.length   > 0) permittedContent.push(groups.join('<br/>'));
```
escattone: Whitespace again: I would prefer only a single space before the `>`.

tests/macros/test-svginfo.js
escattone: It's been a while since I've worked with the test framework, as I forgot that I had already stubbed the `request` package long ago in https://github.com/mdn/kumascript/pull/240! 🤦‍♂️

So your tests were not making real network calls, sorry! I still think it's best to create and populate the stub for `wiki.pageExists` instead of relying on whatever `wiki.pageExists` returns using the un-populated `request` stub.

tests/macros/test-svginfo.js
JeremiePat: Indeed, at that point I need to get into more testing to see what can be standardize (and how) and what cannot and how is it possible to neatly handle special edge case. Those notes are here to remember to have further discussion when we will have more tests.

JeremiePat:
It should be better now.

For `wiki.pageExists` I follow @escattone recommendation as we do not need anything more clever than that in that case.

For the white space, I made the changes even if I'm someone who is very sensitive to vertical alignment but I don't want to fight on this.

PR #694, opened 2018-04-24
adding new existing c to wasm article to webassembly sidebar

chrisdavidmills:
Nice article, added by the google folks: https://developer.mozilla.org/en-US/docs/WebAssembly/existing_C_to_wasm

### DIFF
--- a/macros/WebAssemblySidebar.ejs
+++ b/macros/WebAssemblySidebar.ejs
@@ -8,6 +8,7 @@ var text = mdn.localStringMap({
     'Tutorials' : 'Tutorials',
       'WebAssembly_concepts' : 'WebAssembly concepts',
       'Compiling_to_wasm' : 'Compiling from C/C++ to WebAssembly',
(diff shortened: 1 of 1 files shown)


chrisdavidmills:
I have no idea why this is failing now. It was a really simple addition, and I've checked it thoroughly.

wbamberg:
This macro has a test: https://github.com/mdn/kumascript/blob/master/tests/macros/test-WebAssemblySidebar.js that checks the exact output of the macro, so any change in the sidebar needs a corresponding update to the test. That's why it is failing.

Some of the other tests use jsdom to check output rather than the raw HTML. I'm not sure what we would prefer (I like using jsdom, though, since it helps me to think about what the output of the macro is, rather than just paste the macro output into the test's expected result).

escattone:
@wbamberg 👍 For most cases, I like JSDOM as well, since it restricts focus to the portion of the output that really matters.

PR #697, opened 2018-05-03
add german translations

schlagi123:


SphinxKnight:
@schlagi123 thanks :)

PR #701, opened 2018-05-12
Use the CSS class that indicates "not supported" when the feature is implemented but behind a flag

DavidBruant:
This fixes [bug 1449340](https://bugzilla.mozilla.org/show_bug.cgi?id=1449340)
There is no strong indication that this change is wanted by the MDN team yet, but making the PR nonetheless

This is my first pull request in a long time, so please tell me if i did anything wrong in this PR
I edited the code from the Github UI, so it's not tested at all

### DIFF
--- a/macros/CompatBeta.ejs
+++ b/macros/CompatBeta.ejs
@@ -220,20 +220,27 @@ function getSupportClass(supportInfo) {
     // the first entry should be the most relevant/recent and will be treated as "the truth"
     checkSupport(supportInfo[0].version_added,
                  supportInfo[0].version_removed,
-                 supportInfo[0].partial_implementation);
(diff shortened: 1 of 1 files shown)


macros/CompatBeta.ejs
```diff
@@ -220,20 +220,27 @@ function getSupportClass(supportInfo) {
...
     return 'unknown';
   }
 
-  function checkSupport(added, removed, partial) {
+  function checkSupport(added, removed, partial, flags) {
     if (added === null) {
       cssClass = 'unknown';
     } else if (added) {
-      cssClass = 'yes';
+      if (flags) {
```
saschanaz: Flags sometimes can be enabled by default, which should probably be "yes".

PR #704, opened 2018-05-21
Bug 1387454 - Add AudioContext options

a2sheppy:
Adds the AudioContextOptions dictionary to the interfaces
list for Web Audio API. This is still our only way to
include these in the sidebar, so here it is.

### DIFF
--- a/macros/GroupData.json
+++ b/macros/GroupData.json
@@ -1796,6 +1796,7 @@
                             "AudioBuffer",
                             "AudioBufferSourceNode",
                             "AudioContext",
+                            "AudioContextOptions",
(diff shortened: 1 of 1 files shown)


a2sheppy:
Worth noting there is precedent for including dictionaries in the interfaces list due to not having an official way to include types and the like yet.

a2sheppy:
This needs to be changed to use the new "dictionaries" group instead of putting them into the "interfaces" array.

a2sheppy:
This needs to be changed to use the new "dictionaries" group instead of putting them into the "interfaces" array.

PR #709, opened 2018-06-12
Alphabetize InterfaceData.json.

jpmedley:
I need to add some items for CSS Object Model. It would be nice to have this alphabetized first.


ExE-Boss:
Can we also do mdn/data#358?

PR #720, opened 2018-06-22
bug 1470589: fix CurrentGecko macro

escattone:
This PR fixes a bug in the `CurrentGecko` macro.
- stop using the stale URL (http://svn.mozilla.org/libs/product-details/json/firefox_versions.json) to get Firefox versions (it never responds, so requests eventually timeout)
- the new URL (https://product-details.mozilla.org/1.0/firefox_versions.json) supports "nightly" explicitly (so remove `bumpToNightly` code)
- add tests

Thanks to @jgmize for pointing me to the new URL!

### DIFF
--- a/macros/CurrentGecko.ejs
+++ b/macros/CurrentGecko.ejs
@@ -22,8 +22,6 @@ var channel = $0 || "release";
 channel = channel.toLowerCase();
 
 var versionID;
-var bumpToNightly = false;          // Set to true if we're pulling aurora version but need to increment it
-                                    // since nightly version isn't explicitly listed in the file
 
 switch(channel) {
     case 'release':
@@ -34,7 +32,8 @@ switch(channel) {
         break;
     case 'nightly':
     case 'central':
-        bumpToNightly = true;       // Flag that we'll need to increment, then fall through...
+        versionID = "FIREFOX_NIGHTLY";
+        break;
     case 'aurora':
         versionID = "FIREFOX_AURORA";
         break;
@@ -45,7 +44,7 @@ switch(channel) {
 
 // The URL to pull from
 
-var url = "http://svn.mozilla.org/libs/product-details/json/firefox_versions.json";
+var url = "https://product-details.mozilla.org/1.0/firefox_versions.json";
 
 // Fetch the versions JSON file from MXR.
 
@@ -54,12 +53,6 @@ var versionInfo = mdn.fetchJSONResource(url);
 // We now have a JSON object with the version information, so peel out the value we need
 
 var version = versionInfo[versionID];
-
-if (bumpToNightly) {
-    var major = parseInt(version, 10);  // Get major version number
-    major++;
-    version = major + version.slice(major.toString().length);
-}
 
 // Finally, remove the cruft at the end; we don't need beta numbers etc.
 
--- a/macros/CurrentGecko.ejs
+++ b/macros/CurrentGecko.ejs
@@ -28,18 +28,18 @@ switch(channel) {
         versionID = "LATEST_FIREFOX_VERSION";
         break;
     case 'beta':
+    case 'aurora':
         versionID = "LATEST_FIREFOX_DEVEL_VERSION";
         break;
     case 'nightly':
     case 'central':
         versionID = "FIREFOX_NIGHTLY";
         break;
-    case 'aurora':
-        versionID = "FIREFOX_AURORA";
-        break;
     case 'esr':
         versionID = "FIREFOX_ESR";
         break;
+    default:
+        return 'undefined';
 }
 
 // The URL to pull from
--- a/tests/macros/test-CurrentGecko.js
+++ b/tests/macros/test-CurrentGecko.js
@@ -1,0 +1,57 @@
+/* jshint node: true, mocha: true, esversion: 6 */
+
+var utils = require('./utils'),
+    chai = require('chai'),
+    chaiAsPromised = require('chai-as-promised'),
+    sinon = require('sinon'),
+    assert = chai.assert,
+    itMacro = utils.itMacro,
+    describeMacro = utils.describeMacro,
+    beforeEachMacro = utils.beforeEachMacro;
+
+const ffURL = 'https://product-details.mozilla.org/1.0/firefox_versions.json';
+
+// Let's add "eventually" to assert so we can work with promises.
+chai.use(chaiAsPromised);
+
+describeMacro('CurrentGecko', function () {
+    beforeEachMacro(function (macro) {
+        // Create a test fixture to mock the mdn.fetchJSONResource function.
+        const fetch_stub = sinon.stub();
+        fetch_stub.withArgs(ffURL).returns(
+            {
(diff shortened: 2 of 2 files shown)


connorshea:
Worth mentioning this also happens with the [HTMLSelectElement article](https://developer.mozilla.org/en-US/docs/Web/API/HTMLSelectElement) now.

wbamberg:
> Worth mentioning this also happens with the HTMLSelectElement article now.

Those calls should just be removed. There shouldn't be any calls to this macro (or any macro that calls this macro) anywhere in the open Web docs. But I just tried updating the doc and it seems like my edits didn't get shown.

FYI I made a little chart of all (I think) the macros that call this one:

screen shot 2018-06-25 at 10 37 36 am

tests/macros/test-CurrentGecko.js
```diff
@@ -0,0 +1,57 @@
...
+chai.use(chaiAsPromised);
+
+describeMacro('CurrentGecko', function () {
+    beforeEachMacro(function (macro) {
+        // Create a test fixture to mock the mdn.fetchJSONResource function.
+        const fetch_stub = sinon.stub();
+        fetch_stub.withArgs(ffURL).returns(
+            {
+                'FIREFOX_NIGHTLY': '65.0a1',
+                'FIREFOX_AURORA': '49',
```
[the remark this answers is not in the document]
escattone: No, I changed it to `'49'` as part of making each case in the data unique in order to better detect errors, but I should have left it at an empty string. I'll change it back to an empty string.

escattone:
@wbamberg Are you saying the real fix here is that we should eliminate all calls to the macros shown in your chart (and when that's done, all of those macros as well), rather than fixing the `CurrentGecko` macro?

wbamberg:
@escattone I would love to do that :). At the moment this macro is called pretty extensively (indirectly, through the maze I posted above) in the "Firefox internals" bits of the MDN docs, like the XPCOM reference. These docs are not really maintained and haven't been for years, but I wouldn't like to break them. In the long term (or maybe even the quite short term) it would be great to generate a macro-free "static" version of these docs (as we have also suggested for the archived docs), and then we could remove the macros.

In the immediate term, we should do the simplest thing possible, which I think is what your PR does.

Having said that, this macro should only be called from internals docs, I believe. So in any cases (like https://developer.mozilla.org/en-US/docs/Web/API/HTMLSelectElement, that @connorshea pointed out) where we call it from the open web docs, we can just remove the calls, and that's just good maintenance whether or not the macro actually works :).

tests/macros/test-CurrentGecko.js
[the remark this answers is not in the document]
wbamberg: Aurora has been discontinued (years ago) and replaced with Dev Edition. We could make the macro return the Dev Edition version when given "aurora", but it doesn't really matter.

tests/macros/test-CurrentGecko.js
```diff
@@ -0,0 +1,57 @@
+/* jshint node: true, mocha: true, esversion: 6 */
+
+var utils = require('./utils'),
```
wbamberg: Out of interest, why `var` and not `const` here? We've used `const` in similar cases, e.g.: https://github.com/mdn/kumascript/blob/master/tests/macros/test-toolssidebar.js#L3.

escattone: No good reason at all! 😄 I'll change it to use `const`, thanks @wbamberg!

escattone:
@wbamberg There are currently 7 documents under `/docs/Web` with Kumascript macro errors, all of which I think are related to this issue. In addition to this PR, I'm thinking I should also remove all of the macros you've outlined above from those documents. Does that makes sense to you? Here are the documents:

https://developer.mozilla.org/en-US/docs/Web/API/Document
https://developer.mozilla.org/en-US/docs/Web/API/History_API
https://developer.mozilla.org/en-US/docs/Web/API/HTMLSelectElement
https://developer.mozilla.org/en-US/docs/Web/API/HTMLTextAreaElement
https://developer.mozilla.org/en-US/docs/Web/API/Navigator
https://developer.mozilla.org/en-US/docs/Web/API/ValidityState
https://developer.mozilla.org/en-US/docs/Web/HTML/Element/audio

escattone:
@jwhitlock @wbamberg Thanks for the reviews! I made changes to address your feedback. Would you mind doing a quick final review?

macros/CurrentGecko.ejs
```diff
@@ -22,30 +22,29 @@ var channel = $0 || "release";
...
 var versionID;
-var bumpToNightly = false;          // Set to true if we're pulling aurora version but need to increment it
-                                    // since nightly version isn't explicitly listed in the file
 
 switch(channel) {
     case 'release':
         versionID = "LATEST_FIREFOX_VERSION";
         break;
     case 'beta':
+    case 'aurora':
```
jwhitlock: 👍 I think this is the correct intention for any pages mentioning ``Aurora``.

tests/macros/test-CurrentGecko.js
```diff
@@ -0,0 +1,59 @@
...
+        return assert.eventually.equal(macro.call('CENtral'), '65');
+    });
+    itMacro('Aurora', function (macro) {
+        return assert.eventually.equal(macro.call('aurora'), '61');
+    });
+    itMacro('ESR', function (macro) {
+        return assert.eventually.equal(macro.call('ESR'), '52.8.1');
+    });
+    itMacro('nonsense', function (macro) {
+        return assert.eventually.equal(macro.call('nonsense'), 'undefined');
```
jwhitlock: 👍

wbamberg:
>  Does that makes sense to you?

Yes. As far as I can see these are all through `gecko_minversion*` calls. We should only use BCD for version info. IMO removing the macro calls from pages like these is a good thing.

But it would be easier to see which web docs use these macros once we have eliminated them from the "firefox internals" pages. So my preference for fixing this would be:

1. generate "macro-free" versions of the Firefox internals docs, using the ?raw&macros trick
2. see how many calls to the macros are still left (i.e. which open web docs call them)
3. remove the calls from the open web docs
4. remove the macros from the KS repo

> @wbamberg what tool did you use to create the graph? It looks sharp.

A highly manual combination of `grep -r` and OmniGraffle.

macros/CurrentGecko.ejs
```diff
@@ -55,12 +54,6 @@ var versionInfo = mdn.fetchJSONResource(url);
 
 var version = versionInfo[versionID];
 
-if (bumpToNightly) {
-    var major = parseInt(version, 10);  // Get major version number
-    major++;
-    version = major + version.slice(major.toString().length);
```
ExE-Boss: **Fact:** This had a subtle bug where the original version was `9` (or `99`), which would cause `major++` to increment it to `10` (or `100`), which would cause `major.toString().length` to return a value `1` higher than what was intended, which would cause the `version` string to contain an incorrect value.

Glad that that’s been fixed.

PR #724, opened 2018-06-28
Latest 2 CSS RECs

dansoper:


wbamberg (commented):
I think the change to `REC` is correct but as I understand things we want to URLs to point to the latest draft, so we want to keep https://drafts.csswg.org/css-ui-3/ as the URL. Paging @chrisdavidmills for confirmation.

chrisdavidmills:
@wbamberg Yup, correct.

PR #729, opened 2018-06-30
Add quicklinks sidebar for the Games zone

wbamberg:
This is part of the zone removal work.

It's not ready to merge because it only has the en-US version, and it has no tests. But I'd like feedback on whether this is a sane organization of the sidebar.

I've been pretty conservative but have changed things a bit because quicklinks sidebars can't do multiple levels of indentation.

In particular I've changed the "Techniques" section, renaming it "Basic techniques" and splitting out the 3D games and games controls bits into their own sections. (tbh that section in the existing Games sidebar is a bit of a mess anyway)

### DIFF
--- a/macros/GamesSidebar.ejs
+++ b/macros/GamesSidebar.ejs
@@ -1,0 +1,149 @@
+<%
+
+const locale = env.locale;
+
+const baseURL = `/${locale}/docs/Games/`;
+const webURL = `/${locale}/docs/Web/`;
+const appsURL = `/${locale}/docs/Web/Apps/`;
+
+const text = mdn.localStringMap({
+  "en-US": {
+    "Introduction": "Introduction",
+      "Anatomy": "Anatomy",
+      "Examples": "Examples",
+    "APIs_for_game_development": "APIs for game development",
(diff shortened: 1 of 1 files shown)


wbamberg:
I've changed "Basic Techniques"->"Techniques", and added localizations. Still needs tests, so not ready to merge.

@SphinxKnight , if you were able to take a quick look at the localizations, at least the fr ones, to make sure I haven't mangled anything, that would be really helpful.

wbamberg:
This is ready for review now.

SphinxKnight:
@wbamberg regarding `macros/GamesSidebar.ejs` my editor is warning me about the duplicated key for WebVR (cf. L28 & L46 for `en-US`) on the `text` object . Thanks to the indentation, I can see it's part of two different "submenus" but the object won't (?) "carry" the duplicated key.

I think renaming the key for the second occurrence should work.

Regarding the translations, I've tweaked some existing French ones that needed minor work.

wbamberg:
Thanks @SphinxKnight , good catch on the duplicated key.

PR #730, opened 2018-07-04
Removing duplicate Introducing JavaScript objects menu entry

chrisdavidmills:
In the main JavaScript sidebar (e.g. see https://developer.mozilla.org/en-US/docs/Web/JavaScript), the "Introducing JavaScript Objects" article appears twice, one below "Complete beginners" and once below "Intermediate". This PR removes the latter.

### DIFF
--- a/macros/JsSidebar.ejs
+++ b/macros/JsSidebar.ejs
@@ -34,8 +34,8 @@ var text = mdn.localStringMap({
     'Basics': 'JavaScript basics',
     'First_steps': 'JavaScript first steps',
     'Building_blocks': 'JavaScript building blocks',
-    'Intermediate': 'Intermediate',
     'Introducing_objects': 'Introducing JavaScript objects',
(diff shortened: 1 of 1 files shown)


macros/JsSidebar.ejs
```diff
@@ -34,8 +34,8 @@ var text = mdn.localStringMap({
     'Basics': 'JavaScript basics',
     'First_steps': 'JavaScript first steps',
     'Building_blocks': 'JavaScript building blocks',
-    'Intermediate': 'Intermediate',
     'Introducing_objects': 'Introducing JavaScript objects',
+    'Intermediate': 'Intermediate',
```
wbamberg: I'm not sure why this switches the order for en-US but not for the other locales.

wbamberg (changes requested):
This looks fine. Nit-picking, but I think the strings dictionary should keep the same order for all locales.

PR #746, opened 2018-07-21
Add Japanese translation

WhiteHawk-taka:
Add Japanese translation to AddonSidebar.ejs
Do you have a better idea?

### DIFF
--- a/macros/AddonSidebar.ejs
+++ b/macros/AddonSidebar.ejs
@@ -222,7 +222,7 @@ var embeddedWebExtensions = mdn.localString({
     "ja"    : "Embedded WebExtensions"
 });
 
-var comparisonWithTheAdd-onSDK = mdn.localString({
+var comparisonWithTheAddonSDK = mdn.localString({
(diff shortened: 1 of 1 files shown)


wbamberg:
Thanks for this!

Most of the sidebars adopt a slightly different approach, in which all strings are kept in a single dictionary. See for example https://github.com/mdn/kumascript/blob/master/macros/GamesSidebar.ejs.

As it happens this recent PR: https://github.com/mdn/kumascript/pull/755 uses that method to add French translations to the Add-ons Sidebar. Perhaps it would be best to wait until PR 755 is merged (which should be very soon), then reorganize this one to add the Japanese translations, using the same approach? What do you think?

I'm sorry to make you redo work here, and thanks again for your contribution.

WhiteHawk-taka:
I'll create new PR after #755 is merged.
Thank you @wbamberg .

wbamberg:
@WhiteHawk-taka , https://github.com/mdn/kumascript/pull/755 is merged now, if you are interested in adding Japanese translations on top of that :).

PR #755, opened 2018-07-31
Localization of WebExtensions Sidebar

etiennewan:


EtienneWan:
:wave:
Sorry for all the fixups for typos and missing quotes, maybe squash them ?
This is unfinished, sorry for the inconvenience, but I needed to push this ASAP.

Travis has been failing only on `make lint-macros`  with error `Unexpected token (326:11) in macros/AddonSidebar.ejs`, I haven't found a way to fix this

SphinxKnight:
@EtienneWan ping me whenever you are done :) Thanks for the PR

macros/AddonSidebar.ejs
```diff
@@ -12,145 +12,320 @@ function currentPageIsUnder(root) {
   return "closed";
 }
 
+var text = mdn.localStringMap({
+  'en-US': {
```
wbamberg: We are horribly inconsistent throughout this code in single versus double quotes, but we should at least try to be consistent within each macro.

I think single quotes is better, so would you mind fixing up lines 1-13 of this file to use single quotes?

I see that we also use double quotes for HTML attributes. Perhaps we should leave those ones, that seems to be what we use on MDN.

macros/AddonSidebar.ejs
```diff
@@ -12,145 +12,320 @@ function currentPageIsUnder(root) {
...
+    'WebExtensions/Alternative_distribution_options/Sideloading_add-ons': 'For sideloading',
+    'WebExtensions/Alternative_distribution_options/Add-ons_for_desktop_apps': 'For desktop apps',
+    'WebExtensions/Alternative_distribution_options/Add-ons_in_the_enterprise': 'For an enterprise',
+    'Community_and_Support': 'Community and Support',
+    'Channels': 'Channels',
+    'Blog': 'Add-ons blog',
+    'Forums': 'Add-on forums',
+    '#Contact_us': 'Contact us'
+  },
+  'fr-FR': {
```
wbamberg: I know almost nothing about the way Kuma handles locales, but I think this needs to be just "fr".

If I try your macro, on, say, "http://llocalhost:8000/fr-FR/docs/Web/CSS", then I get redirected to "http://localhost:8000/fr/docs/Web/CSS" and see the en-US strings (I guess this is used as the fallback.

If I change line 102 to just "fr" then I see the French strings.

macros/AddonSidebar.ejs
```diff
@@ -12,145 +12,320 @@ function currentPageIsUnder(root) {
...
+    'WebExtensions/Alternative_distribution_options/Sideloading_add-ons': 'Pour installation manuelle',
+    'WebExtensions/Alternative_distribution_options/Add-ons_for_desktop_apps': 'Pour applications de bureau',
+    'WebExtensions/Alternative_distribution_options/Add-ons_in_the_enterprise': 'Pour une entreprise',
+    'Community_and_Support': 'Communauté et Support',
+    'Channels': 'Canaux de discussions',
+    'Blog': 'Blog extensions',
+    'Forums': 'Forum extensions',
+    '#Contact_us': 'Nous contacter'
+  }
+})
```
wbamberg: End with a semicolon please.

macros/AddonSidebar.ejs
EtienneWan: Fixed, and I agree, it is better to have consistency within a file.

macros/AddonSidebar.ejs
EtienneWan: Fixed :+1:

macros/AddonSidebar.ejs
EtienneWan: Let's blame the linter for that :slightly_smiling_face: fixed :+1:

PR #760, opened 2018-08-09
removing caching compiled webassembly modules article from sidebar

chrisdavidmills:
As per https://bugzilla.mozilla.org/show_bug.cgi?id=1469395


wbamberg:
This fails for the same reason as https://github.com/mdn/kumascript/pull/694.

PR #766, opened 2018-08-16
EmbedInteractiveExample: Add ``?origin=<site URL>`` when not in production

jwhitlock:
Add a query string parameter ``?origin=http://localhost:8000`` when ``EmbedInteractiveExample`` detects that we aren't on production, based on the page URL.  This will communicate to the interactive example code that it should use a different origin for ``postMessage`` calls. See https://github.com/mdn/interactive-examples/pull/1112.

This adds a new helper ``mdn.siteURL()`` which may be useful for other places where we've hard-coded ``developer.mozilla.org`` in macros.

I was also annoyed by extra spaces in the class declaration, so I refactored how the class was constructed, and added some tests.


jwhitlock:
Force-push to fix linting

PR #768, opened 2018-08-18
Fixing markup issue (all article is <strong>)

SphinxKnight:
As detected in https://discourse.mozilla.org/t/why-is-all-normal-text-bold-now/31038
This fixes markup issues introduced by commit 282615b (cf. [L385](https://github.com/mdn/kumascript/blame/master/macros/AddonSidebar.ejs#L385) )


SphinxKnight:
Linking with #769

Issue #769, opened 2018-08-18
All pages using `AddonSidebar` display content within `<strong>`


SphinxKnight:
Creating an issue to replicate https://discourse.mozilla.org/t/why-is-all-normal-text-bold-now/31038 here.

ExE-Boss:
I also opened #771 to fix a remaining `<a>` and `<strong>` DOM hierarchy issue.

SphinxKnight:
@wbamberg yes, thanks for the reminder :)

PR #771, opened 2018-08-20
fix(macros.AddonSidebar): Correct broken <a> and <strong> DOM hierarchy

ExE-Boss:
Follow‑up to #768 and #755.

See also #769.

review?(@SphinxKnight)


SphinxKnight:
Thanks @ExE-Boss & @escattone  :)

PR #775, opened 2018-08-23
macros: introduce WebDriverSidebar

andreastt:
Currently contains a list of commands to be used under
https://developer.mozilla.org/en-US/docs/Web/WebDriver.

### DIFF
--- a/macros/WebDriverSidebar.ejs
+++ b/macros/WebDriverSidebar.ejs
@@ -1,0 +1,150 @@
+<section class=Quick_links id=Quick_Links>
+<ol>
+  <li><a href=/<%=locale%>/docs/Web/WebDriver><strong><em>WebDriver</em></strong></a></li>
+
+  <li class=toggle>
+    <details open>
+      <summary>Sessions</summary>
+
+      <ol>
+        <li><a href=/<%=locale%>/docs/Web/WebDriver/Commands/NewSession>New Session</a>
+        <li><a href=/<%=locale%>/docs/Web/WebDriver/Commands/DeleteSession>Delete Session</a>
+        <li><a href=/<%=locale%>/docs/Web/WebDriver/Commands/Status>Status</a>
+        <li><a href=/<%=locale%>/docs/Web/WebDriver/Commands/GetTimeouts>Get Timeouts</a>
+        <li><a href=/<%=locale%>/docs/Web/WebDriver/Commands/SetTimeouts>Set Timeouts</a>
+      </ol>
+    </details>
+
+  <li class=toggle>
+    <details open>
+      <summary>Navigation</summary>
+
+      <ol>
+        <li><a href=/<%=locale%>/docs/Web/WebDriver/Commands/NavigateTo>Navigate To</a>
+        <li><a href=/<%=locale%>/docs/Web/WebDriver/Commands/GetCurrentURL>Get Current URL</a>
+        <li><a href=/<%=locale%>/docs/Web/WebDriver/Commands/Back>Back</a>
+        <li><a href=/<%=locale%>/docs/Web/WebDriver/Commands/Forward>Forward</a>
+        <li><a href=/<%=locale%>/docs/Web/WebDriver/Commands/Refresh>Refresh</a>
+        <li><a href=/<%=locale%>/docs/Web/WebDriver/Commands/GetTitle>Get Title</a>
+      </ol>
+    </details>
+
+  <li class=toggle>
+    <details open>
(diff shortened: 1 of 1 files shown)


macros/WebDriverSidebar.ejs
```diff
@@ -0,0 +1,110 @@
+<%
```
ExE-Boss: The `<% … %>` section contains JavaScript code to execute, everything outside it is HTML.

You did this in the exact opposite way to how EJS (Embedded JavaScript templates) work.

ExE-Boss (changes requested):

macros/WebDriverSidebar.ejs
ExE-Boss: Also, you can run the tests without docker by running `node run.js`, and then, while the server is running, run `./node_modules/.bin/jshint --show-non-errors lib tests` and `./node_modules/.bin/mocha tests` in a separate console window.

andreastt:
@ExE-Boss Thanks for the pointers, what you said makes obvious sense.

And sorry it took me a while to fixup this patch.

ExE-Boss (commented):
This should be converted to use [`<details>`](https://developer.mozilla.org/docs/Web/HTML/Element/details) (see&nbsp;https://github.com/mdn/kumascript/pull/789).

andreastt:
@wbamberg Is there anything I can do to help writing tests for this sidebar addition?

wbamberg:
@andreastt , I might be mistaken but it looks to me like the issue that @ExE-Boss identified in https://github.com/mdn/kumascript/pull/775#pullrequestreview-150349838 still needs to be fixed.

## Setup

(I don't know how much of this you know, apologies if you know it all already!)

To test a macro locally (both manual testing and writing an automated test) you'd first need to install a local Kuma. There are docs here: https://kuma.readthedocs.io/en/latest/installation.html. You need to do everything up to and including "Visit the homepage".

For manual testing you'll also need to edit pages in your local Kuma, so you'll need to create a user. The easiest way to do that is to [create an admin user](https://kuma.readthedocs.io/en/latest/installation.html#create-an-admin-user).

Then to develop KumaScript macros the best resource is the [KumaScript README](https://github.com/mdn/kumascript/blob/a69bbb3822b6fc9d1f20e8124256a0af93847528/README.md):  in particular [Updating macros](https://github.com/mdn/kumascript/blob/a69bbb3822b6fc9d1f20e8124256a0af93847528/README.md#updating-macros).

## Writing macros and tests

For this macro I would start by adapting an existing sidebar such as https://github.com/mdn/kumascript/blob/a69bbb3822b6fc9d1f20e8124256a0af93847528/macros/ToolsSidebar.ejs. Yours can be simpler though because you don't have to include other translations, although it would be good to make it easy for others to add them later.

For manual testing: assuming you've added the macro under kuma/kumascript/macros:

* restart your local Kuma
* visit http://localhost:8000/admin/login/ and log in as admin so you can edit pages
* open any page (say http://localhost:8000/en-US/docs/Web/HTML) and click the edit button
* add the macro call at the top of the page, like {{WebDriverSidebar}}
* save the page, and see if it looks all right.

For automated testing: the ToolsSidebar macro has a (quite basic, but better than nothing) test here: https://github.com/mdn/kumascript/blob/a69bbb3822b6fc9d1f20e8124256a0af93847528/tests/macros/test-toolssidebar.js and you could adapt that.

It always takes me a while to remember how to run new tests, and this isn't very clearly documented in the KS README, but I believe it's like:

* add the new test script under kumascript/tests/macros
* commit the new test script
* rebuild KumaScript using (assuming you are in the kuma/kumascript directory) `cd ..; make build-kumascript; cd kumascript`
* run the tests using `make test-macros`

I hope that helps. If you get problems try asking in #mdndev, but probably not at the weekend :).

macros/WebDriverSidebar.ejs
```diff
@@ -0,0 +1,150 @@
+<section class=Quick_links id=Quick_Links>
+<ol>
+  <li><a href=/<%=locale%>/docs/Web/WebDriver><strong><em>WebDriver</em></strong></a></li>
```
ExE-Boss: `locale` is `undefined`.

It is `env.locale` that contains what you need.

Also, this sidebar macro isn’t localisable.

ExE-Boss (changes requested):
https://github.com/mdn/kumascript/pull/775#discussion_r213499857 was addressed, but the sidebar macro still has issues.

PR #773, opened 2018-08-22
fix(macros.embed): Make most Embed* macros mostly XSS‑proof

ExE-Boss:
This renders most `Embed*` macros mostly XSS‑proof (not that it’s particularly necessary as MDN already blocks some XSS attacks¹).

¹ [MDN doesn’t have](https://observatory.mozilla.org/analyze/developer.mozilla.org) any [CSP headers](https://infosec.mozilla.org/guidelines/web_security#content-security-policy) (see [bug 948151](https://bugzil.la/948151)).

---

Follow‑up to #766

review?(@jwhitlock, @escattone)

PR #778, opened 2018-08-23
Make EmbedInteractiveExamples mostly XSS‑proof

jwhitlock:
This is the ``{{EmbedInteractiveExamples}}`` changes from #773, with tests.

### DIFF
--- a/tests/macros/test-EmbedInteractiveExample.js
+++ b/tests/macros/test-EmbedInteractiveExample.js
@@ -55,7 +55,7 @@ describeMacro('EmbedInteractiveExample', function () {
         macro.ctx.env.url = 'http://localhost:8000/en-US/docs/Web/HTTP/headers';
         return assert.eventually.equal(
             macro.call('pages/http/headers.html?foo=bar'),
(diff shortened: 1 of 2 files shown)


tests/macros/test-EmbedInteractiveExample.js
```diff
@@ -55,7 +55,7 @@ describeMacro('EmbedInteractiveExample', function () {
         macro.ctx.env.url = 'http://localhost:8000/en-US/docs/Web/HTTP/headers';
         return assert.eventually.equal(
             macro.call('pages/http/headers.html?foo=bar'),
-            `<iframe class="interactive" width="100%" height="250" frameborder="0" src="https://interactive-examples.mdn.mozilla.net/pages/http/headers.html?foo=bar&origin=http://localhost:8000" title="MDN Web Docs Interactive Example"></iframe>`
+            `<iframe class="interactive" width="100%" height="250" frameborder="0" src="https://interactive-examples.mdn.mozilla.net/pages/http/headers.html?foo=bar&amp;origin=http://localhost:8000" title="MDN Web Docs Interactive Example"></iframe>`
```
jwhitlock: 👍 this fixes an ampersand ``&`` escaping issue in the ``src`` attribute. I'm not sure if any live URLs include an ampersand, but now it will actually work 😁

PR #782, opened 2018-08-25
Update Japanese translation of CSS_key_concepts

mfuji09:
Update translation based on recent English version.
Correct error on the French translation.


SphinxKnight (approved):
Thanks for the fix regarding fr.
Looks good to me.

PR #786, opened 2018-08-28
EmbedLiveSample: Avoid XSS, and add allow= parameter

jwhitlock:
Refactor ``{{EmbedLiveSample}}`` to avoid some cross-site scripting (XSS) vectors, including some before-and-after tests. Kuma should bleach these away, but I'd like to start making these templates more XSS-aware, and @ExE-Boss triggered my imagination with PR #773.

Add a seventh optional parameter 😮 that sets an ``allow="<features>"`` attribute for the ``<iframe>``, so that live samples using restricted features can continue working in Chrome and Safari (see [bug 1482159](https://bugzilla.mozilla.org/show_bug.cgi?id=1482159)).

This builds on and includes PR #777 and PR #785, which may be easier to review first.

### DIFF
--- a/tests/macros/test-EmbedLiveSample.js
+++ b/tests/macros/test-EmbedLiveSample.js
@@ -59,6 +59,17 @@ describeMacro('EmbedLiveSample', function () {
             '</iframe>'
         );
     });
+    itMacro('One argument: XSS attempt (success)', function (macro) {
+        macro.ctx.env.url = 'https://developer.mozilla.org/en-US/docs/Web/HTML/Element/figure';
+        macro.ctx.env.revision_id = 1397983;
+        return assert.eventually.equal(
+            macro.call('"><script>alert("XSS");</script>'),
+            '<iframe class="live-sample-frame sample-code-frame"' +
+            ' id="frame_"><script>alert("XSS");</script>" frameborder="0"' +
+            ' src="https://mdn.mozillademos.org/en-US/docs/Web/HTML/Element/figure$samples/%22%3E%3Cscript%3Ealert(%22XSS%22);%3C/script%3E?revision=1397983">' +
+            '</iframe>'
+        );
+    });
     itMacro('Two arguments: ID, width', function (macro) {
         macro.ctx.env.url = 'https://developer.mozilla.org/en-US/docs/Web/CSS/border-top-width';
         macro.ctx.env.revision_id = 1352086;
@@ -67,6 +78,18 @@ describeMacro('EmbedLiveSample', function () {
             '<iframe class="live-sample-frame sample-code-frame"' +
             ' id="frame_Example" frameborder="0"' +
             ' width="100%"' +
+            ' src="https://mdn.mozillademos.org/en-US/docs/Web/CSS/border-top-width$samples/Example?revision=1352086">' +
+            '</iframe>'
+        );
+    });
+    itMacro('Two arguments: ID, XSS attempt (success)', function (macro) {
+        macro.ctx.env.url = 'https://developer.mozilla.org/en-US/docs/Web/CSS/border-top-width';
+        macro.ctx.env.revision_id = 1352086;
+        return assert.eventually.equal(
+            macro.call('Example', '"><script>alert("XSS");</script>'),
+            '<iframe class="live-sample-frame sample-code-frame"' +
+            ' id="frame_Example" frameborder="0"' +
+            ' width=""><script>alert("XSS");</script>"' +
             ' src="https://mdn.mozillademos.org/en-US/docs/Web/CSS/border-top-width$samples/Example?revision=1352086">' +
             '</iframe>'
         );
@@ -107,6 +130,18 @@ describeMacro('EmbedLiveSample', function () {
             '</iframe>'
         );
     });
+    itMacro('Three arguments: ID, width, XSS attempt (success)', function (macro) {
+        macro.ctx.env.url = 'https://developer.mozilla.org/en-US/docs/Web/HTML/Element/figure';
+        macro.ctx.env.revision_id = 1397983;
+        return assert.eventually.equal(
+            macro.call("Images", "100%", '"><script>alert("XSS");</script>'),
+            '<iframe class="live-sample-frame sample-code-frame"' +
+            ' id="frame_Images" frameborder="0"' +
+            ' width="100%" height=""><script>alert("XSS");</script>"' +
+            ' src="https://mdn.mozillademos.org/en-US/docs/Web/HTML/Element/figure$samples/Images?revision=1397983">' +
+            '</iframe>'
+        );
+    });
     itMacro('Four arguments: ID, width, height, ""', function (macro) {
         macro.ctx.env.url = 'https://developer.mozilla.org/en-US/docs/Web/CSS/::before';
         macro.ctx.env.revision_id = 1392665;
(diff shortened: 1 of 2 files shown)


jwhitlock:
PR #785 merged, rebased on master.

tests/macros/test-EmbedLiveSample.js
```diff
@@ -59,6 +59,18 @@ describeMacro('EmbedLiveSample', function () {
             '</iframe>'
         );
     });
+    itMacro('One argument: XSS attempt (failed)', function (macro) {
+        macro.ctx.env.live_samples = {'base_url': 'https://mdn.mozillademos.org'};
```
escattone: Now that you have this in the `beforeEachMacro` call above, it can be removed.

tests/macros/test-EmbedLiveSample.js
```diff
@@ -71,6 +83,19 @@ describeMacro('EmbedLiveSample', function () {
             '</iframe>'
         );
     });
+    itMacro('Two arguments: ID, XSS attempt (failed)', function (macro) {
+        macro.ctx.env.live_samples = {'base_url': 'https://mdn.mozillademos.org'};
```
escattone: Same as above.

tests/macros/test-EmbedLiveSample.js
```diff
@@ -107,6 +132,19 @@ describeMacro('EmbedLiveSample', function () {
             '</iframe>'
         );
     });
+    itMacro('Three arguments: ID, width, XSS attempt (failed)', function (macro) {
+        macro.ctx.env.live_samples = {'base_url': 'https://mdn.mozillademos.org'};
```
escattone: Same as above.

tests/macros/test-EmbedLiveSample.js
```diff
@@ -138,6 +176,26 @@ describeMacro('EmbedLiveSample', function () {
             '</iframe></td></tr></tbody></table>'
         );
     });
+    itMacro('Four arguments: ID, width, height, XSS attempt (failed)', function (macro) {
+        macro.ctx.env.live_samples = {'base_url': 'https://mdn.mozillademos.org'};
```
escattone: Same as above.

tests/macros/test-EmbedLiveSample.js
```diff
@@ -172,6 +230,18 @@ describeMacro('EmbedLiveSample', function () {
             '</iframe>'
         );
     });
+    itMacro('Five arguments: ID, "", "", "", XSS Attempt (failed)', function (macro) {
+        macro.ctx.env.live_samples = {'base_url': 'https://mdn.mozillademos.org'};
```
escattone: Same as above.

tests/macros/test-EmbedLiveSample.js
```diff
@@ -184,4 +254,45 @@ describeMacro('EmbedLiveSample', function () {
             '</iframe>'
         );
     });
+    itMacro('Six arguments: ID, width, height, "", "", XSS attempt (failed)', function (macro) {
+        macro.ctx.env.live_samples = {'base_url': 'https://mdn.mozillademos.org'};
```
escattone: Same as above.

tests/macros/test-EmbedLiveSample.js
```diff
@@ -184,4 +254,45 @@ describeMacro('EmbedLiveSample', function () {
...
+            macro.call("sampleNone",100,50,"","", '"><script>alert("XSS");</script>'),
+            '<iframe class="live-sample-frame &#34;&gt;&lt;script&gt;alert(&#34;XSS&#34;);&lt;/script&gt;"' +
+            ' id="frame_sampleNone" frameborder="0"' +
+            ' width="100" height="50"' +
+            ' src="https://mdn.mozillademos.org/en-US/docs/Web/CSS/-moz-appearance$samples/sampleNone?revision=1402877">' +
+            '</iframe>'
+        );
+    });
+    itMacro('Seven arguments: ID, width, height, "", "", "", features', function (macro) {
+        macro.ctx.env.live_samples = {'base_url': 'https://mdn.mozillademos.org'};
```
escattone: Same as above.

tests/macros/test-EmbedLiveSample.js
```diff
@@ -184,4 +254,45 @@ describeMacro('EmbedLiveSample', function () {
...
+            '<iframe class="live-sample-frame sample-code-frame"' +
+            ' id="frame_Example_Constraint_exerciser" frameborder="0"' +
+            ' width="650" height="800"' +
+            ' src="https://mdn.mozillademos.org/en-US/docs/Web/API/Media_Streams_API/Constraints$samples/Example_Constraint_exerciser?revision=1401022"' +
+            ' allow="video; microphone">' +
+            '</iframe>'
+        );
+    });
+    itMacro('Seven arguments: ID, width, height, "", "", "", XSS Attempt (failed)', function (macro) {
+        macro.ctx.env.live_samples = {'base_url': 'https://mdn.mozillademos.org'};
```
escattone: Same as above.

tests/macros/test-EmbedLiveSample.js
jwhitlock: 🤦‍♂️ of course, good suggestion.

jwhitlock:
Rebased to remove per-test setting of ``macro.ctx.env.live_samples`` like it never happened...

jwhitlock:
Oops, lost change from `XSS attempt (success)` to `XSS attempt (failed)`, one more rebase coming

jwhitlock:
> This is a step in the right direction to help prevent stuff like this from ever getting through.

@ExE-Boss, I should have replied to you 9 days ago, and maybe this should be documented somewhere other than a PR comment, but this makes for a good first draft.

MDN pages are user-submitted content, and so we don't trust them. We use a HTML5-aware library [bleach](https://github.com/mozilla/bleach) that supports an allow-list for tags, attributes, and styles (see [kuma/wiki/constants.py](https://github.com/mozilla/kuma/blob/master/kuma/wiki/constants.py)). MDN wiki content goes through bleaching and other content manipulation before the "rendered" version is displayed to users (see  [kuma/wiki/content.py](https://github.com/mozilla/kuma/blob/master/kuma/wiki/content.py)). 

This keeps users safe, but causes some other issues. The "bleached" content is not fed back to the "raw" content, so users can get confused why the source looks different than the output. Some processes do not preserve attribute order, resulting in large diffs for small changes around tool updates. It could be better, but it is also battle tested over years, so we avoid radical changes.

KumaScript macro output is treated as unsafe content, and goes through the same bleaching process:

https://github.com/mozilla/kuma/blob/6e7fc6489836b9824fa40c65bf61c50ad4c67b87/kuma/wiki/kumascript.py#L213-L217

You'd have to find a weakness in the whole bleaching system in order to get an attack through to visitors. [Mozilla/Add-ons/WebExtensions$revision/1396309](https://developer.mozilla.org/en-US/docs/Mozilla/Add-ons/WebExtensions$revision/1396309) was a weird attempt to add a browser-based coin miner to MDN, embedded as a table. Even if the author had formatted it correctly, the ``<script>`` elements would have been escaped or removed, neutering it.

See the [developer.mozilla.org:Security topic](https://bugzilla.mozilla.org/buglist.cgi?list_id=14318051&resolution=---&resolution=FIXED&resolution=INVALID&resolution=WONTFIX&resolution=INACTIVE&resolution=DUPLICATE&resolution=WORKSFORME&resolution=INCOMPLETE&resolution=SUPPORT&resolution=EXPIRED&resolution=MOVED&query_format=advanced&component=Security&product=developer.mozilla.org) in Bugzilla for the long, painful history of successful XSS attacks and their mitigations. And, if you find something, submit it with the "Security issue" flag, and you might show up on the [Hall of Fame](https://www.mozilla.org/en-US/security/bug-bounty/web-hall-of-fame/). However, MDN is not [one of the sites](https://www.mozilla.org/en-US/security/bug-bounty/web-eligible-sites/) for bounty payouts, so do it for the love, not the money.

We're thinking of how to evolve KumaScript for the future, and some options are a web-accessible KumaScript API or even client-side KumaScript, to fix preview, or to allow alternate contribution paths. Part of that will be making macros safer, so that the output can be trusted to run on user's browsers without Kuma-based bleaching.  We have a long way to go, and haven't decided if we should start.

PR #794, opened 2018-09-01
Add Japanese translation to Compat.ejs

mfuji09:
Add Japanese translation to Compat.ejs


mfuji09:
I have resolved the conflict.

PR #795, opened 2018-09-03
adding Media Queries Level 5 to SpecData

chrisdavidmills:
To support Rachel's work on https://github.com/mdn/sprints/issues/428

### DIFF
--- a/macros/SpecData.json
+++ b/macros/SpecData.json
@@ -414,6 +414,11 @@
     "url": "https://drafts.csswg.org/css-writing-modes-4/",
     "status": "CR"
   },
+  "CSS5 Media Queries": {
+    "name": "Media Queries Level&nbsp;5",
+    "url": "https://drafts.csswg.org/mediaqueries-5/",
+    "status": "ED"
+  },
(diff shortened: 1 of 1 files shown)


macros/SpecData.json
```diff
@@ -414,6 +414,11 @@
     "url": "https://drafts.csswg.org/css-writing-modes-4/",
     "status": "CR"
   },
+  "CSS5 Media Queries": {
+    "name": "Media Queries Level&nbsp;5",
```
wbamberg: Is there a reason we use `&nbsp;` instead of a space?

chrisdavidmills: I am not sure. I just did it like this to match the other spec data. But we probably don't need the `&nbsp;`.

Unfortunately I don't have this branch on my machine anymore, so it comes up as merging from "unknown repository". Could we merge this one first and then I'll fix the `&nbsp;`s afterwards, or do you want me to submit a new PR covering the whole lot?

wbamberg: Sure. I just wondered if there was a reason for this.

PR #796, opened 2018-09-03
Adding CSS Scrollbars Spec

rachelandrew:
So that we can create a property page for `scrollbar-width` https://bugzilla.mozilla.org/show_bug.cgi?id=1475033


chrisdavidmills:
This looks fine, r+

Issue #751, opened 2018-07-27
Intermittent errors in {{bug}}


wbamberg:
I've been making lots of edits to MDN pages lately, and am often seeing macro errors from the [{{bug}}](https://github.com/mdn/kumascript/blob/master/macros/bug.ejs) macro:

<img width="527" alt="screen shot 2018-07-27 at 10 33 03 am" src="https://user-images.githubusercontent.com/432915/43337354-7bdb8330-9188-11e8-8491-6c9e0dcf2ce2.png">

These are intermittent but seem very common when a lot of pages get updated. I suspect this is because the {{bug}} macro [uses the Bugzilla API](https://github.com/mdn/kumascript/blob/master/macros/bug.ejs#L29-L30), and is being rate-limited.

It uses the Bugzilla API to replace duplicated bugs with the bugs that they duplicate. For example, if I write `{{bug("1411526")}}`, then because [bug 1411526](https://bugzilla.mozilla.org/show_bug.cgi?id=1411526) is duped to [bug 1381992](https://bugzilla.mozilla.org/show_bug.cgi?id=1381992), then the HTML output is: a link to bug 1381992.

Personally I don't think this dupe-walking is very useful, and it would be better to remove it, and this would remove the dependency on the Bugzilla API and fix these intermittent errors.

wbamberg:
> From a brief reading of the bug macro, it appears that the loop de-duplicates bugs, but also reads the description and status. We'd lose this as well if we stopped calling the API.

I'd be interested in knowing how many times this macro is called with a "type" of `table` or `summary` (not often I would guess), as outside those cases this info just goes in a tooltip, and I struggle to see how the value that offers is worth the complexity of the code or the dependency on the Bugzilla API.

> I'd love to see some tests to go along with it.

Yes, I think the PR needs tests (although it does look good).

ExE-Boss:
> > I'd love to see some tests to go along with it.
>
> Yes, I think the PR needs tests (although it does look good).

I don’t really know how I should write a test for that though.

emceeaich:
Use a mock for the BMO API. 

https://medium.freecodecamp.org/how-to-mock-requests-for-unit-testing-in-node-bb5d7865814a

a2sheppy:
The table and summary types are used in extremely limited cases. Probably only on meta-doc style pages and possibly the documentation status pages.

emceeaich:
[...]

PR #801, opened 2018-09-06
fix(macros): Make the {{bug}} macro fail‑safe

ExE-Boss:
See #751, note that this doesn’t resolve that, as intermittent failures will still occur, they just won’t cause scripting errors to happen and will produce a slightly better error message.

review?(@wbamberg, @jwhitlock)

### DIFF
--- a/macros/bug.ejs
+++ b/macros/bug.ejs
@@ -26,36 +26,47 @@ var url;
 var bugDetail = [];
 while (bugDupe != null) {
     bugNumber = bugDupe;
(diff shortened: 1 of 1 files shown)


a2sheppy:
I have a concern about this: by changing the macro to generate output without an error, we lose the ability to know that the page needs to be rerendered. I would like to find some way for this to alert us to that; otherwise, we will wind up with pages with these error messages in their HTML that go undetected.

PR #803, opened 2018-09-06
Keyboard Lock and Keyboard Map

jpmedley:


jpmedley:
I've got somebody working on correcting that. Won't don't have to merge now.

jpmedley:
PTAL

It has working links now.

jpmedley:
@Elchi3 These links have worked for a while now. Is there anything else I need to do?

PR #808, opened 2018-09-09
fix(macros.compat): Modernize legacy {{Compat*}} macros

### LINKED ISSUE #799: {{CompatGeckoDesktop}} macro uses zone vanity URLs
URLs should be updated from https://developer.mozilla.org/en-US/Firefox to developer.mozilla.org/en-US/docs/Mozilla/Firefox, to reflect the changes from [bug 1462475](https://bugzilla.mozilla.org/show_bug.cgi?id=1462475).

https://github.com/mdn/kumascript/blob/b566c863c733b7ee3757f65bb5c6e32967ad80e1/macros/CompatGeckoDesktop.ejs#L137

We're using this macro [less](https://developer.mozilla.org/en-US/search?locale=en-US&kumascript_macros=CompatGeckoDesktop&topic=none) with the BCD project, but it will continue being used on Events pages, like https://developer.mozilla.org/en-US/docs/Web/Events/addstream, while we decide if and how to migrate the data.

A macro test for ``CompatGeckoDesktop`` would be nice as well.

ExE-Boss:
Resolves #799

review?(@jwhitlock)


wbamberg:
Sorry, I know this was to fix a big, but I don't think we should spend any time on these legacy macros, unless it's removing calls to them so we can delete them.

PR #809, opened 2018-09-10
Update display spec

rachelandrew:
Spec is a Candidate Recommendation as of 28 August.


macros/SpecData.json
```diff
@@ -1579,4 +1579,4 @@
     "url": "https://www.w3.org/TR/xpath-31/",
     "status": "REC"
   }
-}
+}
```
ExE-Boss: There should be a new‑line here.

PR #815, opened 2018-09-19
add Korean for glossary

cs09g:


alattalatta:
Can this get reviewed? Glossary isn't really functioning at this moment and code itself looks OK.

cs09g:
I'm not sure but it doesn't show proper glossary for current language.
e.g. https://developer.mozilla.org/ko/docs/Gecko_DOM_Reference/%EC%86%8C%EA%B0%9C
Mouseover the `DOM` keyword, it displays english glossary though there's korean traslation for glossary `DOM`(https://developer.mozilla.org/ko/docs/Glossary/DOM).

It works with Japanese language so I wonder why it's not working with Korean language.

cs09g:
@SphinxKnight Is this right way to add macro for Korean language? Then may I send more PRs for other macro files?

SphinxKnight:
@cs09g yes this is quite right. If you plan sending other PRs, please try to make one for each file. It's easier to review/merge and minimizes the risk of conflict :)

PR #888, opened 2018-09-27
adding the new Rust to WASM article to the WebAssembly sidebar

chrisdavidmills:
https://developer.mozilla.org/en-US/docs/WebAssembly/Rust_to_wasm


chrisdavidmills:
OK, let's try this. I forgot we had tests ;-)

PR #919, opened 2018-10-02
Addition of the development tools page to the Firefox workflow section

rebloor:


macros/AddonSidebar.ejs
```diff
@@ -355,8 +356,9 @@ var text = mdn.localStringMap({
     </li>
     <li class="toggle">
       <details>
-        <summary><%=text['WebExtensions#Firefox_workflow']%></summary>
+        <summary><a href="<%=baseURL%>WebExtensions#Firefox_workflow"><%=text['WebExtensions#Firefox_workflow']%></a></summary>
```
wbamberg: As also commented in https://github.com/mdn/kumascript/pull/921: I don't think we should be doing this.

rebloor: Sorry, removed

PR #923, opened 2018-10-04
Add TTML spec data

germain-gg:
It looks like the TTML specification data was not specific anywhere in that document.

https://www.w3.org/blog/news/archives/7344


wbamberg (commented):
Thanks for this PR!

As I understand things, we don't use dated URLs like https://www.w3.org/TR/2018/PR-ttml1-20181004/ in SpecData. There are a couple of other options:

1) reference the latest published version: https://www.w3.org/TR/ttml1/
2) reference the latest draft: https://w3c.github.io/ttml1/index.html

As I understand it our policy is to reference the latest draft, (i.e. (2) above). Also it seems like the status should be `REC` not `PR`?

But paging @chrisdavidmills for a definitive answer.

PR #937, opened 2018-10-12
rm HTMLMainQuickLinks

wbamberg:
Bizarrely, this sidebar was only on one en-US page, https://developer.mozilla.org/en-US/docs/Web/HTML/Reference. I've updated that page to use HTMLSidebar.ejs instead, and done the same for all locales that called it. It can't be worse.

Link to macro: https://github.com/mdn/kumascript/blob/master/macros/HTMLMainQuickLinks.ejs

Search results for references from pages: https://developer.mozilla.org/en-US/search?kumascript_macros=HTMLMainQuickLinks&locale=%2A&topic=none

Search results for internal references: https://github.com/mdn/kumascript/search?utf8=%E2%9C%93&q=HTMLMainQuickLinks&type=


ExE-Boss:
This was apparently meant to be the sidebar macro used on the main HTML (and HTML/Reference) pages, but it’s not a great sidebar macro, at all.

PR #939, opened 2018-10-12
Browser extensions sidebar update - API differences article

rebloor:
Adding "The differences between extensions API implementations" to the porting section.


macros/AddonSidebar.ejs
```diff
@@ -58,6 +58,7 @@ var text = mdn.localStringMap({
     'WebExtensions/Embedded_WebExtensions': 'Embedded WebExtensions',
     'WebExtensions/Comparison_with_the_Add-on_SDK': 'Comparison with the Add-on SDK',
     'WebExtensions/Comparison_with_XUL_XPCOM_extensions': 'Comparison with XUL/XPCOM extensions',
+    'WebExtensions/differences_between_extensions_API_implementations': 'Extension API differences',
```
ExE-Boss: This should be <q>Differences between API implementations</q> located at [WebExtensions/differences_between_API_implementations](https://developer.mozilla.org/docs/Mozilla/Add-ons/WebExtensions/differences_between_API_implementations).

rebloor: Updated

ExE-Boss: The URL still points to the old location.

rebloor: Sorry, URL fixed

rebloor:
@wbamberg Thanks changes, hopefully, made as needed. This article is intended for the new extension workshop site, so is separate from the existing one.

PR #940, opened 2018-10-12
bug 1483072: post-cutover Kumascript clean-up

escattone:
The title says "post-cutover" cleanup, but there's nothing in this PR that has to wait until after the cut-over. The only condition is that this depends on https://github.com/mozilla/kuma/pull/5044 being merged for the tests to pass.
- change the default `IMAGE_PREFIX` within the `Makefile`
-  just use the default `IMAGE_PREFIX` within the `Jenkinsfile`
- update all relevant files to reflect that `quay.io/mozmar` has become `mdnwebdocs`
- anticipates the upcoming cut-over from MozMEAO to MozIT by updating a macro test to reflect that the stage attachments domain has changed from `stage-files.mdn.moz.works` to `files-stage.mdn.mozit.cloud` (this doesn't really matter, but it's just to be thorough)

### DIFF
--- a/Jenkinsfile
+++ b/Jenkinsfile
@@ -76,76 +76,74 @@ node {
             setGitEnvironmentVariables()
         }
     }
-    // TODO: After cutover to IT-owned cluster, make "mdnwebdocs" the default
-    //       value for IMAGE_PREFIX in the Makefile, and remove this code.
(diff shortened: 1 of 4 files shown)


jwhitlock:
Travis is happy now that https://github.com/mozilla/kuma/pull/5044 is merged, but MozMEAO Jenkins will be sad when it tries to push images to dockerhub.

PR #945, opened 2018-10-15
Browser extensions sidebar update - Monetization article

rebloor:
Adding "Make money from browser extension" to the "Publishing at-arms" menu


rebloor:
Thanks @wbamberg  I'm afraid I had assumed this was a new way of specifying the menus since I was last involved - presumably it's not possible to include comments in these files?.  I will shortly submit an update with the correct menu definition. In terms of adding the other locales, this article is being parked here before being moved to the new extensions workshop site, so I'm not sure whether adding information for localization here helps.

macros/AddonSidebar.ejs
```diff
@@ -1,4 +1,4 @@
-<%
+If<%
```
wbamberg: Not sure what this `If` is for.

macros/AddonSidebar.ejs
```diff
@@ -408,12 +411,14 @@ var text = mdn.localStringMap({
       <details>
         <summary><%=text['Distribution/Publishing']%></summary>
         <ol>
+          <li><a href="<%=baseURL%>Distribution"><%=text['Distribution/Signing']%></a></li>
```
wbamberg: I think this is not intended? It just seems to duplicate the previous item.

macros/AddonSidebar.ejs
```diff
@@ -408,12 +411,14 @@ var text = mdn.localStringMap({
...
           <li><a href="<%=baseURL%>Distribution"><%=text['Distribution/Signing']%></a></li>
           <li><a href="<%=baseURL%>WebExtensions/Package_your_extension_"><%=text['WebExtensions/Package_your_extension_']%></a></li>
           <li><a href="<%=baseURL%>Distribution/Submitting_an_add-on"><%=text['Distribution/Submitting_an_add-on']%></a></li>
           <li><a href="<%=baseURL%>Source_Code_Submission"><%=text['Source_Code_Submission']%></a></li>
           <li><a href="<%=baseURL%>Distribution/Resources_for_publishers"><%=text['Distribution/Resources_for_publishers']%></a></li>
           <li><a href="<%=baseURL%>Listing"><%=text['Listing']%></a></li>
+          <li><a href="<%=baseURL%>Distribution/Make_money_from_browser_extensions"><%=text['Distribution/Make_money_from_browser_extensions']%>
```
wbamberg: This should close the `<a>` and `<li>` tags

macros/AddonSidebar.ejs
rebloor: Story, must've been a stray copy paste, removed in latest commit

macros/AddonSidebar.ejs
rebloor: Removed in latest commit

macros/AddonSidebar.ejs
rebloor: fixed in the latest commit

PR #830, opened 2018-09-19
Bug 1492018: Prefer `.blockIndicator` for styling block indicators

ExE-Boss:
See [bug 1492018](https://bugzil.la/1492018) for details.

Depends on https://github.com/mozilla/kuma/pull/4979 and ~~https://github.com/mozilla/kuma/pull/4962~~.

review?(@Elchi3, @jwhitlock, @wbamberg, @escattone)

PR #948, opened 2018-10-16
rm es7

wbamberg:
es7 is obsolete and I just removed it from all pages that called it, which were (all except one) non-en-US.

Link to macro: https://github.com/mdn/kumascript/blob/master/macros/es7.ejs

Search results for references from pages: https://developer.mozilla.org/en-US/search?kumascript_macros=es7&locale=%2A&topic=none

Search results for internal references: https://github.com/mdn/kumascript/search?utf8=%E2%9C%93&q=es7&type=


ExE-Boss:
Oh great, now I’ll have to rebase https://github.com/mdn/kumascript/pull/830, again.

Also, https://github.com/mozilla/kuma/pull/4979 removes the `.overheadIndicator[style="background:%239CF49C;"]` rule.

PR #949, opened 2018-10-16
Make deprecatedGeneric wording more accurate

sideshowbarker:
This change refines the wording of the text output by `deprecatedGeneric.ejs` to be more accurate. Specifically, the wording is changed to emphasize that the feature is no longer recommended, and to clarify that the feature might possibly still be kept for compatibility purposes, and that it *may* have been removed from the relevant Web standards, or *may* be in the process of being dropped (instead of unequivocally stating that is has been removed and is in the process of being dropped).

This language makes the warning language appropriate/accurate for cases like `navigator.appName` and `navigator.appCodeName` — for which, without the wording change, the language is inaccurate and misleading.

See related discussion in https://github.com/mdn/browser-compat-data/pull/2965


chrisdavidmills:
Couple of nitpicks that I think we should change to make it read easier:

* "relevant Web standards" -> "relevant web standards"
* "or may be in the process" -> "may be in the process"

sideshowbarker:
> Couple of nitpicks that I think we should change to make it read easier:
> 
> * "relevant Web standards" -> "relevant web standards"
> * "or may be in the process" -> "may be in the process"

Thanks — made those changes in 7866108

chrisdavidmills:
Perfect, thanks @sideshowbarker ! Merged.

PR #943, opened 2018-10-15
adding imsc specs to kumascript

chrisdavidmills:

PR #954, opened 2018-10-19
Add IMSC 1.0.1 and IMSC 1.1

palemieux:
... to support https://developer.mozilla.org/en-US/docs/Related/IMSC


tairt:
@palemieux @chrisdavidmills As far as I can see this is a duplication of https://github.com/mdn/kumascript/pull/943 and can be closed.

PR #955, opened 2018-10-19
APIRef: Item importance support

a2sheppy:
The dictionaries, types, and callbacks arrays in GroupData.json may now, optionally, specify an object rather than a string as the value of an entry. The object's format is:

```
{ "name": "nameOfItem", "importance": "low|medium|high" }
```

This lets one of these items be classified as to how important it is to expose to the user.

The `APIRef` macro now now supports an optional `options` parameter, which is an object like this:

```
{
  includeDictionaries: "yes|no|important",
  includeTypes: "yes|no|important",
  includeCallbacks: "yes|no|important"
}
```

The parameter must be passed in JSON format (so as a string which is escaped JSON style). The default values for all three options is `important`.

Several of the WebRTC dictionaries and types have been updated to use the new values. This both needed to be done and is useful for testing.


chrisdavidmills:
I've got Kuma up and running on my local machine, and I think I've done the necessary to get your branch of KumaScript installed in it.

But this is not working correctly for me. Let's try to figure out why.

I've got the macro call written like this:

`{{apiref("WebRTC","{includeDictionaries: \"yes\",includeTypes: \"yes\",includeCallbacks: \"yes\"}")}}`

And I get no errors, and the following sidebar output on the screen:

screen shot 2018-10-25 at 11 05 25-fullpage

But I get the same thing regardless of whether I put yes, no, or important in for the object property values.

PR #961, opened 2018-10-24
refactor(Macro Massacre): Delete {{id}}

ExE-Boss:
It turns out that this is completely unused, both in pages and in KumaScript macros.

review?(@Elchi3, @wbamberg)


wbamberg:
@ExE-Boss , thanks for the PR!

I was scared to remove this one, because it has such a hard-to-search name. How did you determine that it's not called internally?

ExE-Boss:
I did a regexp search locally for all instances of `/template\((['"])id\1/` and then a search just for `id` to be extra safe.

Also a search here on Github didn’t really reveal any usage either.

PR #973, opened 2018-11-02
moving the accessiblity inspector to be a core tool

chrisdavidmills:
As per a request from Martin Balfanz


macros/ToolsSidebar.ejs
```diff
@@ -13,13 +13,13 @@ const text = mdn.localStringMap({
       "Network_Monitor": "Network Monitor",
       "Performance": "Performance",
       "Responsive_Design_Mode": "Responsive Design Mode",
+      "Accessibility_inspector": "Accessibility inspector",
```
wbamberg: In the other entries we capitalize "inspector".

chrisdavidmills: Good point, thanks @wbamberg. I've fixed this in all cases except for in the actual page slug, as I can't fix that now.

chrisdavidmills:
Dammnit, I was sure I'd done that, but I messed it up. Trying again.

PR #974, opened 2018-11-05
Bug 1493499, add aria roles to notes/warning boxes

schalkneethling:
This is the KumaScript side of https://github.com/mozilla/kuma/pull/5069


ExE-Boss (changes requested):
There are a lot more block and inline indicator macros that need this.

PR #975, opened 2018-11-05
Update SeeCompatTable.ejs

lp177:
Id translated on fr version
https://discourse.mozilla.org/t/fr-link-href-browser-compat-compatibilite-des-navigateurs/33005


ExE-Boss:
This should probably have been merged _before_ I opened https://github.com/mdn/kumascript/pull/983.

PR #977, opened 2018-11-06
Add overview to Intersection Observer API

a2sheppy:
Add "overview" section to Intersection Observer API's record.

### DIFF
--- a/macros/GroupData.json
+++ b/macros/GroupData.json
@@ -970,6 +970,7 @@
             "events":     [ "message" ]
         },
         "Intersection Observer API": {
+            "overview":   [ "Intersection Observer API"],
             "interfaces": [ "IntersectionObserver",
                             "IntersectionObserverEntry" ],
             "methods":    [],


wbamberg:
Can we have tests for GroupData.json?
It's quite easy to make mistakes updating JSON files, but fairly simple to test at least the basics. I added a [test for SpecData.json](https://github.com/mdn/kumascript/blob/master/tests/macros/test-specdata.js) that you could use as a guide.

a2sheppy:
@wbamberg - I agree GroupData.json needs tests. To be honest, I sort of thought they already existed, at least in the form of linting and a schema. Do those not exist?

wbamberg:
> Do those not exist?

I don't know of a schema for GroupData.

ExE-Boss:
There doesn’t seem to be a schema or tests for that.

a2sheppy:
I think actually getting this approved and merged got lost in that discussion that's not directly related to this change.

wbamberg:
Sorry if I wasn't clear: I was asking for GroupData tests as prerequisite of merging this change, it wasn't just an unrelated comment. We could merge it without of course but it would be safer to have tests. There are lots of other open PRs that update GroupData, and having a test would benefit all of them too.

a2sheppy:
@wbamberg -- I agree GroupData tests are needed. I'm not convinced that it's reasonable to arbitrarily choose to block this particular change to GroupData on them. We make GroupData changes all the time without making any of those other updates write the GroupData test suite.

If someone creates those tests, I'd be happy to add tests for new changes, but blocking a single change individually based on the lack of overall tests seems like an ineffective way to try to get those tests created.

I still need to get this in somehow but my attempts to create new tests have failed badly. :)

wbamberg:
Well: I'm probably not the one staying up all night to sort things out when a macro problem breaks thousands of pages, so in that sense it doesn't matter much what I think.

But I'd like to emphasise why I think GroupData tests, in particular, are important. This wasn't a random suggestion.

**1. They're easy to write**
In the comment above I linked to a test I wrote for SpecData ([it's since been renamed](https://github.com/mdn/kumascript/blob/master/tests/macros/specdata.test.js)). If we can describe the required format for GroupData it should be really quick to adapt the SpecData test. If we can't describe the required format for GroupData, we should not be relying on it. I'm pretty sure that we could have done this in the time taken up arguing about it in this thread.

**2. GroupData is used by thousands of pages**
APIRef depends on GroupData, and APIRef is used on 4000 en-US pages. Last year an update to a JSON file that APIRef depended on broke thousands of pages (https://github.com/mdn/kumascript/pull/792 - not GroupData, admittedly, but the same sort of thing). So the cost of mistakes is high.

**3. GroupData gets a lot of PRs**
Regression tests are useful when the code that they test is being changed. We don't get a lot of PRs for most KS macros. Most of the PRs we do get are updates to sidebars and to the JSON files (GroupData and SpecData especially), so having regression tests for these macros is relatively more important.

Also note though that https://github.com/mdn/kumascript/pull/1041 moves GroupData out of KumaScript and into mdn/data, so if/when that goes ahead, this PR will be made redundant.

a2sheppy:
So are we blocking all GroupData changes until tests are added? We haven't been that I'm aware of, but maybe I missed it. I understand why it needs tests, and you're totally right, it does. My only confusion arises from blocking this specific PR on landing tests for GroupData instead of blocking **all** PRs that try to modify GroupData.

a2sheppy:
As for "easy to write" -- I think that depends on your experience and skill set. I have not been able to write a single successfully functioning test to date, despite multiple attempts.

chrisdavidmills:
Bear in mind that we brought this up in Orlando, and the devs kinda agreed that writing tests should be in their domain. Ryan was looking at the possibility of a valid HTML linter...

Maybe worth pining Ryan about this to see if where this got to, and if he can help with writing the tests in general.

escattone:
I haven't had time to further explore the HTML linting for Kumascript (although I did find what seems like a nice [Python-based HTML linter](https://html5lib.readthedocs.io/en/latest/) that I [used for https://github.com/mdn/stumptown-experiment](https://github.com/mdn/stumptown-experiment/blob/c57781831c58196a6126be5d503c2bd9503636ce/code/python/stump/core/prose.py#L58). I'm happy to add that work as well as writing some tests for `GroupData` as part of my tasks for the next sprint

a2sheppy:
@chrisdavidmills -- Oh yeah, that's right. I'd forgotten that that had wound up on their plate.

@escattone -- If you can do that stuff next sprint, that would be most excellent. Would help to unblock a good bit of work. :)

wbamberg:
It might be better (or at least, also be good) to put some time into helping https://github.com/mdn/kumascript/pull/1041 to happen, assuming that's the direction we want to go in.

wbamberg:
@a2sheppy , @chrisdavidmills , since I now have to update GroupData.json for the Web Crypto docs, I will write a test for it.

a2sheppy:
@wbamberg -- Woot! You are most awesome!

a2sheppy:
No spec that I know of. It's not hard to figure out though.

-- 
Eric Shepherd
Senior Technical Writer
Mozilla Developer Network
https://developer.mozilla.org/
Blog: http://www bitstampede.com/
Twitter: https://twitter.com/sheppy

ExE-Boss:
I have put together a schema for GroupData.json in&nbsp;https://github.com/mdn/data/pull/342 based on how it’s used in KumaScript macros.

PR #978, opened 2018-11-06
CSS Scrollbars is at FPWD

rachelandrew:
Can we use status FPWD? Or would you prefer WD? Let me know if you want it changed.


wbamberg:
I think this is failing because of the test I added the other day, that checks the status is one of these: https://github.com/mdn/kumascript/blob/master/tests/macros/test-specdata.js#L9.

I took that list of valid statuses from the list in the spec2 macro (https://github.com/mdn/kumascript/blob/master/macros/spec2.ejs#L26) - basically if `status` is anything else, then that macro will output "Unknown".

So I think we should either:

* add FPWD to the spec2.ejs macro, and to the test-specdata.js test, or
* change this to WD

I don't really have an opinion on which of these is better though. But it is nice to see that the test is running :).

rachelandrew:
changed it to a WD, I don't think there is any real benefit it being flagged as FPWD.

PR #963, opened 2018-10-24
Fix {{Non‑standard_Header}} in preview mode

ExE-Boss:
To preserve `git blame` and history, this should be merged using the **Merge&nbsp;commit** strategy.

review?(@davidflanagan)

PR #981, opened 2018-11-06
Bug 1451261, add class to macros for use in custom styling

schalkneethling:
This adds the class `icon-only-inline` that will be used as a hook for custom styling on the Kuma side. This forms part of https://github.com/mozilla/kuma/pull/5058

### DIFF
--- a/macros/NonStandardBadge.ejs
+++ b/macros/NonStandardBadge.ejs
@@ -44,7 +44,7 @@ var titleAttrValue = mdn.localString({
 var str = "";
 
 if ($0) {
-    %><span title="<%=titleAttrValue%>"><i class="icon-warning-sign"> </i></span><%
+    %><span title="<%=titleAttrValue%>" class="icon-only-inline"><i class="icon-warning-sign"> </i></span><%
 } else {
     str = mdn.localString({
         "ca": "Non-standard",
--- a/macros/ObsoleteBadge.ejs
+++ b/macros/ObsoleteBadge.ejs
@@ -43,7 +43,7 @@ var titleAttrValue = mdn.localString({
 var str = "";
 
 if ($0) {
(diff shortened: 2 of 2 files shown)


jwhitlock:
I'd like this to have tests, and it seems they would be easy to write, but I don't think I can force you to write them.

schalkneethling:
@escattone Can you provide some pointers on writing tests for these changes? There does not seem to be any existing so, will need to know how to start the new tests. Thanks!

escattone:
@schalkneethling The first step is to add a new test file for each of these macros, for example `test-NonStandardBadge.js` and `test-ObsoleteBadge.js`, within the [`tests/macros` directory](https://github.com/mdn/kumascript/tree/master/tests/macros). I think [`kumascript/tests/macros/test-draft.js`](https://github.com/mdn/kumascript/blob/master/tests/macros/test-draft.js) might be a good starting-point/template, but of course you can select another example within the `tests/macros` directory that you feel suits your needs better. Then when you're ready to run your tests:
```
cd ..; make build-kumascript; cd kumascript
make test-macros
make lint-macros
```

I hope that helps get you started, and I'm happy to help beyond that if you have any further questions!

jwhitlock:
https://github.com/mdn/kumascript/pull/963 includes tests for non-standard, so you may be 90% of the way to tests with that one, but unfortunately also renames some macros, so you'd need to be careful that this code still works as desired.

ExE-Boss:
This shouldn’t break #963, as the tests use JSDOM and don’t care about the presence of the `class` attribute.

Also, there are no merge conflicts.

ExE-Boss:
This needs #983 and #984 to be merged, as those icons don’t have the `icon-only-inline` class.

PR #983, opened 2018-11-07
Add tests for Experimental indicator macros

ExE-Boss:
Based on the code from #963.

---

review?(@wbamberg)


ExE-Boss:
review?(@chrisdavidmills)

chrisdavidmills:
As an FYI, I'm not really the best person to review KS tests; I've never really touched that side of the code (although I could learn, given the time)

ExE-Boss:
review?(@wbamberg): This is now ready for&nbsp;review and&nbsp;merging.

macros/ExperimentalBadge.ejs
```diff
@@ -6,82 +6,23 @@
...
-    "pt-BR": "Experimental",
-    "pt-PT": "Experimental",
     "zh-CN": "实验性 API",
-    "zh-TW": "Experimental",
-    "en-US": "Experimental"
 });
-/* ja : OK */
 
-%><%-template("SimpleBadge", [str, "experimentalBadge", titleAttrValue])%><%
+%><%-template("SimpleBadge", [str, "experimental", titleAttrValue])%><%
```
wbamberg: Can you please explain why this change is being made?

macros/SeeCompatTable.ejs
```diff
@@ -1,17 +1,30 @@
...
+    "pt-BR" : ">Verifique a <a href='#Compatibilidade_de_navegadores'>tabela de compatibilidade entre Navegadores</a> cuidadosamente antes de usar essa funcionalidade em produção.",
+    "ru"    : "Так как спецификация этой технологии ещё не стабилизировалась, смотрите <a href='#Browser_compatibility'>таблицу совместимости</a> по поводу использования в различных браузерах. Также заметьте, что синтаксис и поведение экспериментальной технологии может измениться в будущих версиях браузеров, вслед за изменениями спецификации.",
+    "zh-CN" : "此功能某些浏览器尚在开发中，请参考<a href='#Browser_compatibility'>浏览器兼容性表格</a>以得到在不同浏览器中适合使用的前缀。由于该功能对应的标准文档可能被重新修订，所以在未来版本的浏览器中该功能的语法和行为可能随之改变。",
+    "zh-TW" : "此功能在某些瀏覽器尚在開發中，請參考<a href='#Browser_compatibility'>兼容表格</a>以得到不同瀏覽器用的前輟。",
 });
 
 %><div class="blockIndicator experimental indicator-warning">
-    <p><%-template("ExperimentalBadge", [1])%> <%- str %></p>
+    <p><strong><%- str_title %></strong>
+    <br/><%- str_long %></p>
```
wbamberg: Can you please explain why this change is being made?

wbamberg (commented):
It's great to see more tests. But this PR seems to make some other changes too, that aren't explained. I think I'd prefer if this *only* added tests, unless the tests have found bugs that need to be fixed. And then I'd like an explanation of why these changes are needed.

If the changes are just refactoring, I would prefer them to be in a separate PR.

Thanks @ExE-Boss !

macros/ExperimentalBadge.ejs
ExE-Boss: Because one of the `*Badge` classes doesn’t get the icon applied, but I don’t remember which one.

macros/SeeCompatTable.ejs
ExE-Boss: It is to match all the other `*_Header` macros which store the title and description texts in separate variables.

PR #984, opened 2018-11-07
Add tests for remaining Deprecated indicator macros

ExE-Boss:
Based on the code from https://github.com/mdn/kumascript/pull/963.

~~Also adds the `icon-only-inline` class from #981.~~

---

Depends on #1145 and #1106

review?(@davidflanagan)


davidflanagan:
I'm unqualified to review this, since I don't know what the macros are used for. I do find it surprising that you're making code changes to macros that have "Deprecated" in their names. That doesn't seem wise: if they're deprecated, we should be focused on stablity and getting rid of existing uses not on improving the implementation. It is also misleading to give the PR a title that just says it is adding tests when it also makes significant changes to non-test code.

I do not know who has the time and the expertese to review this.

ExE-Boss:
They’re not deprecated, they’re used to insert the 👎 icon into text.

PR #985, opened 2018-11-08
removing all of the moz browser api methods that are being removed in…

chrisdavidmills:
… Firefox 65

As per https://bugzilla.mozilla.org/show_bug.cgi?id=1503070

We might want to wait until 65 actually comes to release, but maybe not.


macros/GroupData.json
```diff
@@ -101,41 +101,28 @@
             "events":     [ "message" ]
         },
         "Browser API": {
-            "guides":     [ { "url":   "/docs/Web/API/Using_the_Browser_API",
+            "guides":     [ { "url":   "/docs/Mozilla/Gecko/Chrome/API/Browser_API/Using",
```
ExE-Boss: I moved this to [`"/docs/Mozilla/Gecko/Chrome/API/Browser_API/Using_the_Browser_API"`](https://developer.mozilla.org/docs/Mozilla/Gecko/Chrome/API/Browser_API/Using_the_Browser_API)

chrisdavidmills: Well, I didn't really want it moved there. I felt the amount of words in the slug was redundant, so I was trying something else.

wbamberg:
Browser API is now removed from GroupData, so I think this can be closed.

PR #989, opened 2018-11-10
Add Turkish (tr) translation

ahmetelgun:


SphinxKnight (changes requested):
Thanks @ahmetelgun, this may seem a bit picky but could you change the files so that the locales code are sorted alphabetically?

SphinxKnight (approved):
Thanks @ahmetelgun :) Looks good to me!

PR #992, opened 2018-11-12
bug xxxx: Enable debugger during test runs

MatonAnthony:
Enable the Node debugger during the test run.

- The debugger is enabled when running the tests with `make test-debug`
- The debugger is exposed on port 9229
- The debugger automatically breaks on the first line of the code

# Next steps:
- [ ] Open or find a bug to link this to.


a2sheppy:
Note to self: @MatonAnthony has another PR coming that will enable tests while running the entire Kuma instance.

PR #996, opened 2018-11-16
Use the name httpRequest for request module

a2sheppy:
This resolves a keyword collision between the module
and the test framework sinon. Without this fix, attempts
to use the request module while testing KumaScript
macros instead calls into the sinon module's request
component, which doesn't work as expected, breaking
tests.

### DIFF
--- a/lib/kumascript/api.js
+++ b/lib/kumascript/api.js
@@ -24,7 +24,7 @@ var url = require('url'),
     async = require('async'),
     Fiber = require('fibers'),
     Future = require('fibers/future'),
-    request = require('request'),
+    httpRequest = require('request'),
     Memcached = require('memcached'),
     ks_errors = require(__dirname + '/errors'),
     ks_utils = require(__dirname + '/utils');
@@ -153,7 +153,7 @@ var APIContext = ks_utils.Class({
     // easily make HTTP requests.
     //
     // TODO: Very permissive. Should there be more restrictions on net access?
-    request: request,
+    request: httpRequest,
 
     // #### zlib
     //


jwhitlock:
I'm not familiar with your issues, but your test should not be making unmocked HTTP requests. I suspect your test needs changes rather than the API.

a2sheppy:
Why not? It seems the best way to test is with real data. The amount of stuff I would have to have built into the test to test this is insane, if I have to avoid allowing the data to be loaded from the JSON files.


--
Eric Shepherd
Senior Technical Writer
MDN Web Docs
https://developer.mozilla.org/

escattone:
@a2sheppy As @jwhitlock said, you should be mocking all network requests. Two important reasons are that:
- it guarantees consistency in the returned data (the data returned from the real network request will vary when updated)
- it avoids the intermittent failures and delays that can occur when making real network requests, which, in turn, introduce false failures that force us to manually interpret failures in our test suite

The macro test framework was written assuming that all network requests would be mocked, and enforces it by swapping out the real `request` npm package with a `sinon` stub (see https://github.com/mdn/kumascript/blob/7ef7f880ad1355b34f3262e1db5e48c13581c904/tests/macros/utils.js#L19).

a2sheppy:
Well huh. I was told not to mock this stuff because it invalidated the test to do so. Now I’m looking at having to replicate vast amounts of data into the test script. I think it would take at least a couple-three sprints to build even the one test I need to add for this then...


--
Eric Shepherd
Senior Technical Writer
MDN Web Docs
https://developer.mozilla.org/

jwhitlock:
@a2sheppy I think at this point you should open a PR with your work in progress and reassign to me or @escattone. It sounds like you've hit the deep end of macro testing too early.

MatonAnthony:
Oh sorry I assumed it was stubbed :(, I didn't understood that you wanted to make a real network call.

wbamberg:
I don't think this PR is going to be merged.

PR #998, opened 2018-11-20
Added a new sub-menu for the Network Monitor section of DevTools docs.

irenesmith:
Added a new sub-menu to the Network Monitor section of the sidebar to provide access to the new pages created when I broke up the extremely long original page.


irenesmith:
I think I properly addressed your comments. Thanks!

irenesmith:
> I've had more thoughts about this. You should remove the slashes from the property/field names, so for example instead of "Network_Monitor/Toolbar", make it "Network_Monitor_Toolbar", or even "Toolbar" (provided this doesn't cause a naming conflict)
> 
> Also, in your list items you are using the property/field values, not the property/field names, so for example currently
> 
> `<li><a href="<=%=baseURL%>Network_Monitor/Throttling"><%=text["Throttling"]%></li>`
> 
> should be
> 
> `<li><a href="<=%=baseURL%>Network_Monitor/Throttling"><%=text["Network_Monitor/Throttling"]%></li>`

I think I have not addressed your requested changes.

irenesmith:
Why has this not been merged? I updated the PR according to the last comments from @chrisdavidmills so I don't understand why it is still sitting here.

chrisdavidmills:
I don't think you did this one:

> I've had more thoughts about this. You should remove the slashes from the property/field names, so for example instead of "Network_Monitor/Toolbar", make it "Network_Monitor_Toolbar", or even "Toolbar" (provided this doesn't cause a naming conflict)

irenesmith:
> 
> 
> I don't think you did this one:
> 
> > I've had more thoughts about this. You should remove the slashes from the property/field names, so for example instead of "Network_Monitor/Toolbar", make it "Network_Monitor_Toolbar", or even "Toolbar" (provided this doesn't cause a naming conflict)

I've made those changes as well.

chrisdavidmills:
> I've made those changes as well.

Ooh, one more thing; you still need to correct the JSON member names in the actual menu list structure. For example <%=text["Network_Monitor/Toolbar"]%> should be <%=text["Network_Monitor_Toolbar"]%>

irenesmith:
@chrisdavidmills I've made all of the changes you requested. What is the remaining problem?

chrisdavidmills:
@irenesmith It wasn't just missing underscores — there are also some words that need removing. Have another look at my last comment.

PR #999, opened 2018-11-21
removing the webgl content from the learning area sidebar, and adding…

chrisdavidmills:
… it to the WebGL sidebar instead

So I decided to remove the WebGL content from the Learning area; see the description on this page for why: https://developer.mozilla.org/en-US/docs/Learn/WebGL

The "learn by example" articles are still useful, so I've moved them into the WebGL section: https://developer.mozilla.org/en-US/docs/Web/API/WebGL_API/By_example

This PR accounts for that move in the sidebars for those docs trees.


chrisdavidmills:
Can I please get a review on this PR, and get it merged? The Learning Area has been sat there with a broken link on it's sidebar for over 2 months now, and it has been reported to me about 30 times. Thanks ;-)

chrisdavidmills:
Thanks @wbamberg, you are a lifesaver.

PR #1000, opened 2018-11-22
fix(macros): Correct {{Draft}} username detection pattern

ExE-Boss:
While using the `{{Draft}}` macro, I found out that my username gets parsed as&nbsp;<code>[ExE](https://developer.mozilla.org/profiles/ExE)&#x2011;Boss</code> instead&nbsp;of&nbsp;<code>[ExE&#x2011;Boss](https://developer.mozilla.org/profiles/ExE-Boss)</code>

This also makes it so that the macro outputs localised URLs.

review?(@davidflanagan, @jwhitlock)


jwhitlock:
Thanks for the update, and for testing this untested feature. I can guess why you are unhappy with the username parsing 😁 

Can you try the [USERNAME_REGEX](https://github.com/mozilla/kuma/blob/master/kuma/users/constants.py#L8) used in Kuma for username validation? It would be good to keep those in sync.

jwhitlock:
I'm not a fan of changing the name from ``draft.ejs`` to ``Draft.ejs``. Most wiki code uses ``{{ draft }}``. Can you change it back?

ExE-Boss:
I prefer changing the name from `draft.ejs` to `Draft.ejs` because it looks better.

PR #1004, opened 2018-11-23
Update locales that use new location in event.ejs

### LINKED ISSUE #1029: {{event}} macro produces links to old URLs in the all loclaes except en-US
The URLs of the documents of events have been moved in the all locales, however {{event}} macro still produces links to old URLs for them, in the all loclaes except en-US.

old: /<locale>/docs/Web/Reference/Events/
new: /<locale>/docs/Web/Events/

mfuji09:
Most of locales have moved to new location (/docs/Web/Events), so corrected to use the list languages which use old locations.
Fixes #1029

### DIFF
--- a/macros/event.ejs
+++ b/macros/event.ejs
@@ -18,7 +18,7 @@ var URL = "/" + lang + "/docs/Web/Events/" + api + $2;
  
 // Locals that don't use the new location yet
 // Remove your local when you have moved your pages
-var filter = ['es', 'ko'];
+let filter = [];
 
 if (filter.indexOf(lang) > -1) { URL = "/" + lang + '/docs/Web/Reference/Events/' + api + $2; }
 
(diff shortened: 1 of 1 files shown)


macros/event.ejs
```diff
@@ -16,11 +16,11 @@ var str = $1 || $0;
  
 var URL = "/" + lang + "/docs/Web/Events/" + api + $2;
  
-// Locals that use the new location
-// Add your local when you have moved your pages
-var filter = ['en-US'];
+// Locals that don't use the new location yet
+// Remove your local when you have moved your pages
+var filter = ['es', 'ko'];
```
ExE-Boss:
```suggestion
let filter = [];
```

ExE-Boss (changes requested):
I went ahead and moved the remaining Spanish and Korean locales to new locations:
- https://developer.mozilla.org/es/docs/Web/Events
- https://developer.mozilla.org/ko/docs/Web/Events

ExE-Boss (approved):
LGTM, r+

ExE-Boss:
Also, it might be a good idea to put `Fixes #1029` into [the first comment](https://github.com/mdn/kumascript/pull/1004#issue-233278299).

PR #1005, opened 2018-11-25
Fix french slug for SVGAttr

a-mt:
The URL for SVG attributes in french is actually [/fr/docs/Web/SVG/Attribute](https://developer.mozilla.org/fr/docs/Web/SVG/Attribute), not Attributs - it shouldn't be localized.

### DIFF
--- a/macros/SVGAttr.ejs
+++ b/macros/SVGAttr.ejs
@@ -2,7 +2,6 @@
 /* one parameter: attribute name */
 var slug = mdn.localString({
     "en-US": "Attribute",
-    "pl"   : "Atrybut"
 });
 
 var URL = "/" + env.locale + "/docs/Web/SVG/" + slug + "/" + $0;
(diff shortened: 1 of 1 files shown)


macros/SVGAttr.ejs
```diff
@@ -3,7 +3,6 @@
 var slug = mdn.localString({
     "en-US": "Attribute",
     "de"   : "Attribut",
```
ExE-Boss: The same applies to `de`: [`/de/docs/Web/SVG/Attribute`](https://developer.mozilla.org/de/docs/Web/SVG/Attribute)

macros/SVGAttr.ejs
```diff
@@ -3,7 +3,6 @@
...
-    "fr"   : "Attributs",
     "pl"   : "Atrybut"
```
ExE-Boss: [`/pl/docs/Web/SVG/Attribute`](https://developer.mozilla.org/pl/docs/Web/SVG/Attribute) isn’t translated and [`Atrybut`](https://developer.mozilla.org/pl/docs/Web/SVG/Atrybut) returns a 404.

PR #1001, opened 2018-11-22
chore(package): Update chai to version 4.2.0

ExE-Boss:
Also fixes the `test-server.js` when run as `./node_modules/.bin/mocha tests/test-server.js`.

review?(@davidflanagan, @jwhitlock)

Blocks [bug&nbsp;1505084](https://bugzilla.mozilla.org/show_bug.cgi?id=1505084)

PR #1008, opened 2018-11-27
Bug 1505084: update KumaScript to use Node 10.13.0

davidflanagan:
This commit just updates the package.json file in the KumaScript repo
and regenerates the corresponding npm-shrinkwrap.js file. Along with the
Node verison update, we update the version of the fibers package because
the old version we were using has a native component that would no longer
compile with Node 10.

The Kuma repo contains the Dockerfile that creates the Docker image
for KumaScript. When Kuma is updated to use this revision of KumaScript
we must update the Dockerfile at the same time or we will have a
version mismatch between the version of node specified in package.json
and the version actually installed in the KumaScript container.

Test Plan:
- tested following the directions at
  https://github.com/davidflanagan/kumascript#updating-the-dockerfile-andor-packagejson-file
- ran kuma locally without problems

### DIFF
--- a/package.json
+++ b/package.json
@@ -16,7 +16,7 @@
     "test": "make test"
   },
   "engines": {
-    "node": "8.10.0"
+    "node": "10.14.2"
   },
   "directories": {
     "lib": "./lib"
@@ -25,7 +25,7 @@
     "async": "2.1.4",
     "ejs": "2.5.7",
     "express": "4.14.0",
-    "fibers": "1.0.15",
+    "fibers": "3.1.1",
     "hirelings": "0.4.3",
     "mdn-browser-compat-data": "0.x",
     "memcached": "2.2.2",
--- a/npm-shrinkwrap.json
+++ b/npm-shrinkwrap.json
@@ -5,11 +5,11 @@
   "requires": true,
   "dependencies": {
     "@newrelic/koa": {
-      "version": "1.0.5",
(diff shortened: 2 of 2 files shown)


ExE-Boss:
See&nbsp;also https://github.com/mdn/kumascript/pull/1001

jwhitlock:
@ExE-Boss PR #1001 will be obsoleted by the move to jest for tests. I prefer that people use the process at https://github.com/mdn/kumascript#updating-the-dockerfile-andor-packagejson-file for ``package.json`` changes, which uses ``npm`` in the Docker container to generate a consistent ``npm-shrinkwrap.json``, regardless of local tools. You could continue using ``pnpm``, and use that process when prepare your PR.

package.json
```diff
@@ -16,7 +16,7 @@
     "test": "make test"
   },
   "engines": {
-    "node": "8.10.0"
+    "node": "10.14.2"
```
jwhitlock: I'd accept ``">=10.14.2 <11"`` to get future versions of node 10, but I'm not sure how often node minor releases break things.

ExE-Boss: Minor node releases almost never break things.

```suggestion
    "node": "^10.14.2"
```
(`^x.y.z` is&nbsp;equivalent to&nbsp;setting `>=x.y.z <(x+1)`)

PR #1032, opened 2019-01-04
bug 1404669: optional cleanup

escattone:
This PR performs some minor cleanup related to the work that was done for  [bug 1404669](https://bugzilla.mozilla.org/show_bug.cgi?id=1404669). None of this cleanup is required to actually fix the bug.

- update the default setting for `document_url_template`  to reflect the addition of `redirect=no`
- the `template_url_template` setting has been obsolete for some time, so it has been removed

### DIFF
--- a/lib/kumascript/conf.js
+++ b/lib/kumascript/conf.js
@@ -39,9 +39,7 @@ var DEFAULT_CONFIG = {
         workerRetries: 3,
         template_root_dir: "macros",
         document_url_template:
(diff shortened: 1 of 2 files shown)


escattone:
@davidflanagan Thanks for the review! Other than as a kind of documentation of what the real value is expected to look like, there really is no benefit to adding `redirect=no` to the default values for the `document_url_template` setting (since they're never used). 😄

PR #1033, opened 2019-01-06
chore: Migrate to ESLint

ExE-Boss:
This migrates from JSHint to ESLint using the configuration from Kuma.

## Blocks:
- #1015

## Todo:
- [x] Fix lint errors

---

review?(@davidflanagan, ~@Elchi3~, @jwhitlock)

PR #1035, opened 2019-01-08
Modernize KumaScript

davidflanagan:
This large patch is a major rewrite of the KumaScript server. It:

- removes the Fiber/Future code and replaces it with async/await
- updates many macros to use await on any function calls that were previously blocking
- removes the hirelings package and macro_worker class because they are no longer needed now that templates are non-blocking
- removes the GET endpoint; all rendering is now done with POST
- removes the complicated (and unused) HTTP caching code
- replaces memcached with an in-memory LRU cache (with a 1-hour expiration) for documents requested by macros
- gets rid of "autorequired" macros and instead makes the MDN-Common and DekiScript-* part of the standard API available to all macros
- removes unused features like the REPL and statsd
- removes dependencies on underscore and other packages
- modernizes the JavaScript code to use classes, let, for/of etc.
- rewrites tests to use Jest
- formats code with Prettier

The rewritten KumaScript server:

- cuts the number of lines of code in half
- reduces external npm dependencies from 15 to 8
- improves rendering speed by more than a factor of 10
- uses about half the CPU load since it does not fork additional processes.


davidflanagan:
This is the same patch as https://github.com/mozilla/kuma/pull/5174, but done against the KumaScript repo instead of moving the new server into Kuma.

davidflanagan:
Hmm. Looks like Travis doesn't know how to test this PR. I'll have to get that working before we can land this, of course.

ExE-Boss:
See also https://github.com/mdn/data/pull/342

jwhitlock:
@davidflanagan I think you should rebase rather than manually merge in changes from master. If you need a code freeze for this change, we may need to advertise that more widely, and see if there are ways to break it into smaller mergable bits.

davidflanagan:
When I first did this PR, I'd forgotten about the macro tests in tests/macros/. I have now added a commit that restores those tests and makes them work. As part of this, I moved them to macros/test/ (because the tests in tests/ use jest and these macro tests are still using mocha).

I thought github would be smart enough to recognize that this was just a move and that they are not new files. But apparently not: it seems to be showing everything under macros/tests/ as a new file. Sorry about that. You certainly don't need to review any of the fixture files. The only significant changes are to macros/tests/util.js, and macros/tests/dekiscript-{page,wiki}.js

davidflanagan:
The Kuma changes required for this patch are in https://github.com/mozilla/kuma/pull/5182

.eslintrc.json
```diff
@@ -0,0 +1,21 @@
...
+    },
+    "extends": [
+        "eslint:recommended",
+        "plugin:jest/recommended",
+        "prettier"
+    ],
+    "parserOptions": {
+        "ecmaVersion": 2018
+    },
+    "rules": {
```
ExE-Boss: Also, please take a&nbsp;look at&nbsp;https://github.com/mdn/kumascript/pull/1033 for some of&nbsp;the&nbsp;rules I’d like to&nbsp;enforce.

src/environment.js
```diff
@@ -0,0 +1,1114 @@
...
+        }
+
+        // Now wait for all the promises to resolve
+        let strings = await Promise.all(parts);
+
+        // Join it all together and return it
+        return strings.join('');
+    },
+
+    StartsWith(str, sub_str) {
```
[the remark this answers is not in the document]
davidflanagan: Not as part of this patch. It is such a big change that ensuring backward compatibliity with the old KumaScript is my highest concern. So I'm not going to make any additional changes to macros now.  

(I agree that it is ridiculous that our macro api includes `strings.StartsWith()`, but it is also harmless...)

src/environment.js
```diff
@@ -0,0 +1,1114 @@
...
+         */
+        function prepareProto(o, binding) {
+            let p = {};
+            for (let [key, value] of Object.entries(o)) {
+                if (binding && typeof value === 'function') {
+                    value = value.bind(binding);
+                }
+                p[key] = value;
+                p[key.toLowerCase()] = value;
+                p[key[0].toUpperCase() + key.slice(1)] = value;
```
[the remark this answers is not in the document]
davidflanagan: It could be implemented with a Proxy. I'm not sure that I agree that it should be, however. It is not like we're planning to broaden the case-insensitive API surface. Rather than doing proxies, I'd rather go find all the non-canonical uses of these APIs and convert them to be canonical.

Again, however, not as part of this patch.

.eslintrc.json
davidflanagan: What I see in #1033 is mostly rules being disabled. But many of those don't need to be disabled since I'm using prettier for code formatting.

I don't have particularly strong feelings about eslint rules, so if you want to make specific arguments for or against specific rules let's talk about that after this lands.

ExE-Boss: I disabled them to&nbsp;allow the&nbsp;build to&nbsp;pass (the&nbsp;rules in&nbsp;#1033 are all copied from Kuma, where they’re all&nbsp;enabled).

davidflanagan:
This most recent commit adds an overview to the README and clarifies and corrects the comments at the top of each .js file.

Makefile
```diff
@@ -38,4 +36,7 @@ shrinkwrap:
 	docker run -it -v ${MOUNT_DIR}\:${APP_DIR} -w / -u root ${IMAGE} \
 	    bash -c "npm shrinkwrap && cp npm-shrinkwrap.json ${APP_DIR}"
 
-.PHONY: run test test-macros lint lint-macros bash shrinkwrap
+src/parser.js: src/parser.pegjs
+	./node_modules/.bin/pegjs src/parser.pegjs
```
jwhitlock: 👍 I think it was an error that we weren't generating `src/parser.js` for the production images, one that's been there for a few years. In general, I don't like checking in derived files, but this one probably changes rarely enough that it won't be an issue.

README.md
```diff
@@ -11,15 +11,85 @@ The KumaScript service takes requests to render raw documents (documents that
...
+  is the clever middleware for doing that.
+
+- _src/config.js_ defines configurable constants for KumaScript such
+  as the port that the server listens on, and the size of the LRU
+  cache. Some of these constants take their values from environment
+  variables.
 
-![KumaScript overview of GET](overview.png)
-
-NOTE: If you'd like to update the diagram above, import `overview.xml` into
```
jwhitlock: I'll need to remember to grab a copy of these files for future docs.

README.md
```diff
@@ -140,7 +214,7 @@ more before you run your local development version of MDN.
 
 ## Setup (Standalone)
 
-* Install [Node.js 8.x](https://nodejs.org/en/download/package-manager/)
+* Install [Node.js 10.x](https://nodejs.org/en/download/package-manager/)
```
jwhitlock: 👍 Good catch!

README.md
```diff
@@ -11,15 +11,85 @@ The KumaScript service takes requests to render raw documents (documents that
...
 fully-rendered document (a document where each of the embedded macros
 has been executed and replaced inline with its output). The
-requests can be either via GET, with the full document path in the URL like
-`GET /docs/en-US/JavaScript/Foo`, or via `POST /docs`, with the actual raw
-content of the document included in the body of the POST. Here's an overview
-of a GET request:
+requests are via the `POST /docs` endpoint, with the raw
+content of the document to be rendered included in the body of the POST.
+
+### Source code
```
jwhitlock: I like this section, it is a good summary of your exploration and rewriting of the server code. However, I expect 99% of the changes will be to macros, so I think it should follow the section on updating macros.

README.md
```diff
@@ -149,20 +223,18 @@ more before you run your local development version of MDN.
...
-        * TODO: Ignore this file.
-* To generate document macro parser (optional):
-    * `./node_modules/.bin/pegjs lib/kumascript/parser.pegjs`
-        * This is not required in dev, but should be done for production.
-        * If `parser.js` is missing, the parser will be built on the fly.
+    * `npm run lint`
+* To generate document macro parser (if parser.pegjs is modified):
+    * `make parser.js`
+    * After regenerating this file, add the line `/* eslint-disable */`
+      to the top of the file.
```
jwhitlock: This rarely changes, and the linter would catch it, but we could automate adding the lint comment:
```
echo "/* eslint-disable */" > src/parser.js
./node_modules/.bin/pegjs -o - src/parser.pegjs >> src/parser.js

File: src/server.js [hunk 9]
```diff
@@ -0,0 +1,276 @@
...
+        // This will raise an error if the macros are not found
+        this.templates = new Templates(config.macrosDirectory);
+
+        // Configure a logger that pipes to the winston logger.
+        if (config.logging) {
+            var logger = morgan(
+                ':remote-addr - - [:date] ":method :url ' +
+                    'HTTP/:http-version" :status :res[content-length] ' +
+                    '":referrer" ":user-agent" :response-time',
+                {
```
escattone: I think we should toss this second argument to the `morgan` constructor, so we'd just end-up using the default stream which is `process.stdout`. The second argument messes up the logging, in that we get the object as a string, like this:
```
{"source":"server","pid":1,"level":"info","message":"::ffff:192.168.80.1 - - [Tue, 15 Jan 2019 20:00:16 GMT] \"GET /macros HTTP/1.1\" 304 - \"-\" \"Mozilla/5.0 (Macintosh; Intel Mac OS X 10.14; rv:64.0) Gecko/20100101 Firefox/64.0\" 102.841"}
```
rather than, without it, the formatted string, like this:
```
::ffff:192.168.48.1 - - [Tue, 15 Jan 2019 18:24:56 GMT] "GET /macros HTTP/1.1" 304 - "-" "Mozilla/5.0 (Macintosh; Intel Mac OS X 10.14; rv:64.0) Gecko/20100101 Firefox/64.0" 3.196
```

That would get us one step closer to dumping `winston` as well.

src/server.js
```diff
@@ -0,0 +1,276 @@
...
+                    }
+                }
+            );
+
+            app.use(logger);
+        }
+
+        // Define a response.log.error() function that allows us to send
+        // KumaScript error messages back to Kuma in HTTP response headers
+        app.use(firelogger({ logger: config.logging ? winston.info : null }));
```
escattone: In my reading of the `firelogger` code, it doesn't use the `logger` attribute of the options object passed into it, unless I'm missing something? I would vote for dumping `winston` here, which I think is the final use of `winston`, so we could dump it as a dependency as well, and simplify to this:
```
app.use(firelogger());
```

Makefile
```diff
@@ -7,29 +7,27 @@ APP_DIR ?= /app
...
-	    /node_modules/.bin/jshint --show-non-errors lib tests
+	  /node_modules/.bin/eslint *.js src tests
 
-lint-macros:
+lint-json:
 	docker run ${DOCKER_RUN_ARGS} ${IMAGE} \
-	    /node_modules/.bin/ejslint "macros/**/*.ejs"
-	docker run ${DOCKER_RUN_ARGS} ${IMAGE} \
-	    /node_modules/.bin/jsonlint-cli "macros/**/*.json"
+	  /node_modules/.bin/jsonlint -q *.json `find src macros tests -name '*.json'`
```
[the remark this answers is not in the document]
davidflanagan: This works:

```
$ find . -name '*.json' -not -path './node_modules/*' -print0 | xargs -n1 -0 ./node_modules/.bin/jsonlint -q
```
Unfortunately, it doesn't print the filename of the file with an error when it fails, so it is sub-optimal. Weird that there is not an obviously better tool on npm. Lots of options for actually validating against a schema, but not an obvious command-line tool for looking for syntax errors.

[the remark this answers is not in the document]
davidflanagan: This is a brand-new repo, but it looks like it would do what we want: lint all json files, ignoring the directories in .gitignore: https://www.npmjs.com/package/json-linter-2

src/environment.js
```diff
@@ -0,0 +1,1118 @@
...
+            }
+            return freeze(p);
+        }
+
+        this.templates = templates;
+        let globals = Object.create(prepareProto(globalsPrototype));
+        globals.__kumascript = globals;
```
escattone: I don't understand the need for this. It seems that within the functions that have been bound to `globals` by `prepareProto`, it should be equally valid to use, for example, `this.env.locale` as `this.__kumascript.env.locale`. What am I missing?

src/server.js
davidflanagan: I suppose I could probably just pass `console.log` instead of `winston.info` to retain the current behavior.

escattone: Oh man, I was looking at the code in https://github.com/lmorchard/node-firelogger, forgetting that the `firelogger` code was brought into this repo (and I think I was the one that did it, lol)! Sorry! If it was useful for you, I think we should keep it.

escattone: Oops, I missed your latest comment. Good point, if you're OK with it, let's just use `console.log` instead of `winston`, and then we can dump the dependency on `winston`.

macros/CSP.ejs
```diff
@@ -1,20 +1,20 @@
...
-var rtlLocales = ['ar', 'he', 'fa'];
-var localStrings = string.deserialize(template("L10n:Common"));
-var URL = "/" + lang + '/docs/Web/HTTP/Headers/Content-Security-Policy/' + directive;
-var anch = '';
+let lang = env.locale;
+let directive = $0;
+let str = $1 || $0;
+let rtlLocales = ['ar', 'he', 'fa'];
+let localStrings = string.deserialize(await template("L10n:Common"));
+let URL = "/" + lang + '/docs/Web/HTTP/Headers/Content-Security-Policy/' + directive;
```
schalkneethling: @davidflanagan As we are using some ES6 goodies, should we not consider using template literals as well? Might it cause problems with the way the string is built?

macros/Compat.ejs
```diff
@@ -37,7 +37,7 @@ var output = '';
 var category = query.split(".")[0];
 var legendItems = new Set(); // entries will be unique
 
-var compatStrings = string.deserialize(template('L10n:CompatTable'));
+var compatStrings = string.deserialize(await template('L10n:CompatTable'));
```
schalkneethling: @davidflanagan Is there a reason we still have some uses of `var` even though in other places they are changed to `let` or `const`?

macros/Embed_text.ejs
```diff
@@ -26,14 +26,14 @@ var m2 = mdn.localString({
 if (brush.indexOf($1) > -1) {
     syntaxe = $1;
 }
-  
-env.files.forEach(function(aFile) {
+
+for(let aFile of env.files) {
```
schalkneethling: Nit: Our code standard here (not properly documented yet ;() requires a space between `for` and the opening parentheses. I believe that is what Prettier does as well? @davidflanagan

macros/Glossary.ejs
```diff
@@ -8,12 +8,12 @@
...
 
-var str = $1 || $0;
-var css = !$2 ? "glossaryLink" : "";
+let str = $1 || $0;
+let css = !$2 ? "glossaryLink" : "";
 
 // Terme location
-var URL_EN = "/en-US/docs/Glossary/";
-var URL    = mdn.localString({
+let URL_EN = "/en-US/docs/Glossary/";
```
schalkneethling: @davidflanagan Should we define these as `const` perhaps?

ExE-Boss: Probably.

macros/CSP.ejs
davidflanagan: We're running in node 10, so all of ES6 is up for grabs if macro authors want to use it. 

For this patch, though, I'm trying to touch the macros as little as possible, so I don't want to add any more changes.

macros/Compat.ejs
davidflanagan: I tried to keep the changes to the macros/ directory minimal. It was not my goal to modernize the JavaScript in the macros.

Obviously, I had to add await (and some asyncs) to the macros. There were also a few cases where I had to change from var to let to get the macros to work correctly.  IIRC the main issue was where we had `var page`. That declaration shadows the page.* API. And for some reason related to moving the DekiScript-page.ejs macro into src/environment.js I needed to change all `var page` declarations into `let page`.  

And I think that sometimes when I was changing one var to a let, I'd go ahead and change the surrounding ones as well.  But there was no intent to do it everywhere (because that can alter the behavior of code, and I didn't want to risk it.)

macros/Embed_text.ejs
davidflanagan: I've always written them without spaces. But prettier inserts them, so we're good on that account. 

If we take Florian at his word, the macros are going to be eliminated, so I don't think it is worth taking the time to make them look nice.  (At least not in this PR!)

macros/Glossary.ejs
davidflanagan: I'd rather not make any more macro changes in this PR, especially since it risks changing the behavior of the macros in subtle ways.

Makefile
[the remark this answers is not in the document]
davidflanagan: Fixed by commit b200111

Makefile
```diff
@@ -38,4 +36,7 @@ shrinkwrap:
 	docker run -it -v ${MOUNT_DIR}\:${APP_DIR} -w / -u root ${IMAGE} \
 	    bash -c "npm shrinkwrap && cp npm-shrinkwrap.json ${APP_DIR}"
 
-.PHONY: run test test-macros lint lint-macros bash shrinkwrap
+src/parser.js: src/parser.pegjs
+	./node_modules/.bin/pegjs src/parser.pegjs
+
+.PHONY: run local-tests test test-macros lint lint-macros bash shrinkwrap
```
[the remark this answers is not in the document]
davidflanagan: Thanks; fixed.

README.md
davidflanagan: Thanks for that suggestion. Done in commit  682f6bd

README.md
davidflanagan: I ended up moving to to the end of the README instead.

Makefile
```diff
@@ -7,29 +7,27 @@ APP_DIR ?= /app
...
+local-tests:
+	npm run test
+	npm run lint
+	npm run lint-json
 
-test-macros:
+test:
 	docker run ${DOCKER_RUN_ARGS} ${IMAGE} \
-	    /node_modules/.bin/mocha --timeout=${TEST_RUN_TIMEOUT} ${TEST_RUN_ARGS} tests/macros
+	  /node_modules/.bin/jest -w1
```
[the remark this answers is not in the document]
escattone: I was thinking that perhaps we don't really care about the code coverage for `firelogger.js` and `parser.js`, so perhaps we should exclude them? Jest has a setting `coveragePathIgnorePatterns` that looks like the one to use if we decide to do that.

src/server.js
davidflanagan: Fixed in af2b87f.

Thanks for pointing this out. I was cargo-culting the old logging code. The logs are much cleaner now!

src/server.js
davidflanagan: Fixed in af2b87f 
Winston is gone!

package.json
```diff
@@ -1,54 +1,43 @@
...
+        "format": "./node_modules/.bin/prettier --write --require-pragma *.js src/*.js tests/*.js tests/macros/*.js"
+    },
+    "engines": {
+        "node": "10.14"
+    },
+    "dependencies": {
+        "ejs": "2.6.1",
+        "express": "4.16",
+        "lru-cache": "5.1.1",
+        "mdn-browser-compat-data": "0.0.59",
```
escattone: For now at least, I think we should keep this at:
```
"mdn-browser-compat-data": "0.x",
```
We did that originally so we could easily add the latest version of `mdn-browser-compat-data` without submitting a Kumascript PR (which happens almost every week on Thursday), and we have a [line in the Kumascript Dockerfile](https://github.com/mozilla/kuma/blob/9589b848ff098cce4c4827fc15668e4bae4d3c5a/docker/images/kumascript/Dockerfile#L29) explicitly for taking advantage of that and updating to the latest.

macros/DekiScript-Page.ejs
```diff
@@ -1,225 +0,0 @@
...
-            }
-        }
-    },
-
-    // If ordered is true, the output is an <ol> instead of <ul>
-    //
-    // Special note: If ordered is true, pages whose locale differs from
-    // the current page's locale are omitted, to work around misplaced
-    // localizations showing up in navigation.
-    subPagesFormatList: function (pages, ordered) {
```
escattone: Nice catch noticing that this isn't used and not including it your new code!

macros/DekiScript-Page.ejs
```diff
@@ -1,225 +0,0 @@
...
-                result = subpages.subpages || [];
-            } else {
-                result = [subpages];
-            }
-        }
-        return result;
-    },
-
-    // If reverse is non-zero, the sort is backward
-    subPagesSort: function (pages, reverse) {
```
escattone: Nice catch noticing that this isn't used and not including it your new code!

macros/DekiScript-String.ejs
```diff
@@ -1,94 +0,0 @@
...
-    EndsWith: function (str, suffix) {
-        str = (''+str);
-        return str.indexOf(suffix, str.length - suffix.length) !== -1;
-    },
- 
-    Contains: function (str, sub_str) {
-        return (''+str).indexOf(sub_str) !== -1;
-    },
-     
-    Compare: function (first, second, ignorecase) {
```
escattone: Nice catch noticing that this isn't used and not including it your new code!

macros/DekiScript-String.ejs
```diff
@@ -1,94 +0,0 @@
...
-        else { return 0; }
-    },
-
-    // http://developer.mindtouch.com/en/docs/DekiScript/Reference/DekiScript_Functions_and_Variables/String.Deserialize
-    Deserialize: function (str) {
-        return JSON.parse(str);        
-    },
-    
-    // http://developer.mindtouch.com/en/docs/DekiScript/Reference/DekiScript_Functions_and_Variables/String.IndexOf
-    IndexOf: function (first, second) {
```
escattone: Nice catch noticing that this isn't used and not including it your new code!

macros/DekiScript-Uri.ejs
```diff
@@ -1,32 +0,0 @@
-<% module.exports = buildAPI({
- 
-    // Decompose a URI into its component parts. (Obsoletes uri.parse)
-    // [See also](http://developer.mindtouch.com/en/docs/DekiScript/Reference/DekiScript_Functions_and_Variables/Uri.Parts)
-    parts: function (str) {
```
escattone: Nice catch noticing that this isn't used and not including it your new code! Well, I should say it isn't used by anything that is actually executed, because it appears to be used within `kumascript/tests/fixtures/server/macros/xulelem.ejs` but that is never executed and seems to be only used as a filename in the list of macros returned for testing purposes. I suppose we could remove `kumascript/tests/fixtures/server/macros/xulelem.ejs` (and update `kumascript/tests/fixtures/server/macros-expected.json` accordingly)?

package.json
davidflanagan: Thank you for noticing that! I pinned it to that specific version because that was what was in use when I generated my test corpus with the old KumaScript, and in order to compare compat tables, I needed the same version. But that was just supposed to be temporary!

macros/DekiScript-Web.ejs
```diff
@@ -1,35 +0,0 @@
...
-        if (target) {
-            out.push(' target="' + kuma.htmlEscape(target) + '"');
-        }
-        out.push('>', kuma.htmlEscape(text || uri), '</a>');
-        return out.join('');
-    },
-    
-    // Convert source to HTML by optionally fetching and parsing it.
-    // [See also](http://developer.mindtouch.com/en/docs/DekiScript/Reference/DekiScript_Functions_and_Variables/Web.Html)
-    html: function (str) {
```
escattone: LOL, what a useful function! 😄 Nice catch noticing that this isn't used and not including it your new code!

Makefile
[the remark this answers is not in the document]
davidflanagan: Added the test-coverage targets in commit da5e381 and also tweaked the code and tests to improve coverage.

macros/DekiScript-Wiki.ejs
```diff
@@ -1,379 +0,0 @@
-<%  var wiki = module.exports = buildAPI({
-    /*
-    * Given a MediaWiki path attempt to update it to a Kuma path
-    * This is not a DekiScript function
-    */
-    kumaPath: function(in_path) {
```
escattone: Nice catch noticing that this isn't used and not including it your new code!

Makefile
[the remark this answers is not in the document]
davidflanagan: Good idea about coveragePathIgnorePatterns. I've used it in a126c15 to exclude src/parser.js
Leaving firelogger.js in the coverage for now, though.

macros/DekiScript-Page.ejs
davidflanagan: Thanks. I tried to minimize the API surface whereever I could.

macros/DekiScript-Wiki.ejs
```diff
@@ -1,379 +0,0 @@
...
-            var mdn = require_macro('MDN:Common');
-            return mdn.cacheFnIgnoreCacheControl(key, 3600, regenerate);
-        } else {
-            return cacheFn(key, 3600, regenerate);
-        }
-    },
-
-    // Readable shortcut to wiki.page with optional parameters
-    // and ignore_cache_control=true
-    pageIgnoreCacheControl: function (path, section, revision, show, heading) {
```
escattone: Nice catch noticing that this isn't used and not including it your new code!

macros/DekiScript-Wiki.ejs
```diff
@@ -1,379 +0,0 @@
...
-    //
-    //  level:  the heading level (1-6)
-    //  title:  the heading's text title, as displayed to the user
-    //
-    // Parameters:
-    //
-    //  path    The path of the page whose headings should be returned
-    //  depth   The maximum header level to return
-
-    getHeadings: function(path, depth) {
```
escattone: Nice catch noticing that this isn't used and not including it your new code!

macros/DekiScript-Wiki.ejs
```diff
@@ -1,379 +0,0 @@
...
-        var out = parts.protocol + '//' + parts.host + wiki.preparePath(path);
-        if (query) {
-            out += '?' + query;
-        }
-        return out;
-    },
-
-    // A link to the wiki page that you will add, that will be in a different language.
-    // [See also](http://developer.mindtouch.com/en/docs/DekiScript/Reference/Wiki_Functions_and_Variables/Wiki.Languages)
-    languages: function (langs) {
```
escattone: Nice catch noticing that this isn't used and not including it your new code!

macros/DekiScript-Uri.ejs
davidflanagan: I'll dig into that and try to clean it up. It appears with different capitalizations in other fixtures, as well.

macros/DocStatus.ejs
```diff
@@ -11,9 +11,6 @@
 //         e.g. { "cleanup": {"title": "Clean-up", "link": "#Cleanup_on_standard_built-in_ECMAScript_objects", "counter": 142 }}
 //  $6  -  (optional) Tags that all pages should have. (OR logic)
 
-/*** Module loading ***/
-var MDN = require_macro("MDN:Common");
```
escattone: Nice to see this removed!

macros/HTMLRefTable.ejs
```diff
@@ -91,78 +91,86 @@ function hasCommonTag(A, B) {
...
+            tagLink      : (t !== 'heading_elements'
+                               ? await template("HTMLelement", [t])
+                               : (await Promise.all([
+                                   template("HTMLelement", ['h1']),
+                                   template("HTMLelement", ['h2']),
+                                   template("HTMLelement", ['h3']),
+                                   template("HTMLelement", ['h4']),
+                                   template("HTMLelement", ['h5']),
+                                   template("HTMLelement", ['h6'])
+                               ])).join(', ')),
```
escattone: Nicely done!

package.json
davidflanagan: Fixed in 5b40b6e

macros/MDN-Common.ejs
```diff
@@ -1,467 +0,0 @@
...
-            '</a>'
-        ].join('');
-    },
-    
-    /**
-     * Given a set of names and a corresponding list of values, apply HTML
-     * escaping to each of the values and return an object with the results
-     * associated with the names.
-     */
-    htmlEscapeArgs: function (names, args) {
```
escattone: Nice catch not to carry-forward this unused function!

macros/MDN-Common.ejs
```diff
@@ -1,467 +0,0 @@
...
-            b += c;
-        }
-        return b.replace(/(<([^>]+)>)/ig, "");
-    },
-    
-    /**
-     * #### fetchCompatTableJSON
-     * Sets proper headers and makes a request to get compat table data
-     */
-    fetchCompatTableJSON: function(url) {
```
escattone: Nice catch not to carry-forward this unused function!

macros/MDN-Common.ejs
```diff
@@ -1,467 +0,0 @@
...
-        });
-        f.wait();
-        if (err_result) { throw err_result; }
-        return result;
-    },
-    
-    // #### memcacheSet(key, value, timeout)
-    //
-    // Store a value in memcache with the given key and timeout (in seconds)
-    memcacheSet: function (key, value, tmout) {
```
escattone: 👍 Another unused function is gone!

macros/MDN-Common.ejs
```diff
@@ -1,467 +0,0 @@
...
-        });
-        f.wait();
-        if (err_result) { throw err_result; }
-        return result;
-    },
-    
-    // #### memcacheGet(key)
-    //
-    // Fetch the value for a key from memcache
-    memcacheGet: function (key) {
```
escattone: 👍 And another!

macros/MDN-Common.ejs
```diff
@@ -1,467 +0,0 @@
...
-        };
-        if (opts.ignore_cache_control) {
-            return MDN.cacheFnIgnoreCacheControl(opts.cache_key, opts.cache_timeout, to_cache);
-        } else {
-            return cacheFn(opts.cache_key, opts.cache_timeout, to_cache);
-        }
-    },
-    
-    /* Returns the appropriate string (usually a comma+space) for usage in a enumeration list for the current locale */
-    listSeparator: function() {
```
escattone: 👍 Another useless function bites the dust! :)

macros/MDN-Common.ejs
```diff
@@ -1,467 +0,0 @@
...
-        return mdn.localString(listSeparators);
-    },
-    
-    /*
-     * FILE READING
-     *
-     * Utilities to read various files.
-     */
-
-    CSVToArray: function( strData, strDelimiter ) {
```
escattone: And another!

macros/MDN-Common.ejs
```diff
@@ -1,467 +0,0 @@
...
-        // Now that we have our value string, let's add
-        // it to the data array.
-        arrData[ arrData.length - 1 ].push( strMatchedValue );
-      }
-    
-      // Return the parsed data.
-      return( arrData );
-    },
-    
-    loadArrayFromCSV: function(url, separator) {
```
escattone: And another.

macros/HTMLRefTable.ejs
ExE-Boss: Shouldn’t it be `template("HTMLElement")`?

macros/XULMethInc.ejs
```diff
@@ -29,11 +29,10 @@ var destOldType = lang + '/docs/' + 'XUL:' + methodName + ':' + api;
 var destNewType = lang + '/docs/' + 'Mozilla/Tech/XUL/' + methodName + '/' + api;
 
 if (wiki.pageExists(destNewType)) { /* the page exists from the deki age */
-    %><%-template("page", [destNewType])%><%
+    %><%-await template("page", [destNewType])%><%
 } else if(wiki.pageExists(destOldType)) { /* the page exists from the mediawiki age */
-    %><%-template("page", [destOldType])%><%
+    %><%-await template("page", [destOldType])%><%
 } else { /* the page doesn't exist */
-    log.debug("Neither exists");
```
escattone: 😄 Nice catch!

src/server.js
escattone: LOL, love that phrase "cargo-culting", which I hadn't heard before! 😆

macros/HTMLRefTable.ejs
davidflanagan: Template names are case-insensitive, so I just kept the capitalization that was already in use in this file.

macros/HTMLRefTable.ejs
```diff
@@ -91,78 +91,86 @@ function hasCommonTag(A, B) {
...
             tagName      : t,
-            
-            // We need to special case the "Heading elements" article which stand for all hn elements.
-            tagLink      : t !== 'heading_elements' ? template("HTMLelement", [t])
-                         : ['h1','h2','h3','h4','h5','h6'].map(function (tg) { return template("HTMLelement", [tg]); }).join(', '),
-            
+            // We need to special case the "Heading elements" article
+            // which stand for all hn elements.
+            tagLink      : (t !== 'heading_elements'
+                               ? await template("HTMLelement", [t])
```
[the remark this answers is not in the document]
davidflanagan: I'd rather not make cosmetic changes like this in the macro files in this PR.

macros/DekiScript-Uri.ejs
davidflanagan: Resolved in 5b40b6e
I deleted the xulelem.ejs and two other macros in that fixtures directory that were not being used by tests except for the /macros/ endpoint test. Those were the three files that used this old Uri.Parts() function.

src/environment.js
davidflanagan: Fixed in 39dac67

While fixing up this file, it occurred to me that I could move the big blobs of API functions (mdnPrototype, wikiPrototype, stringPrototype, etc.) into their own modules, maybe in a new  src/api/ directory. Then I could require() those into environment.js. It would make the file not so long, and would separate the code with poor test coverage from the code with good test coverage.

I like the idea, and have added a TODO comment about it. With your approval, I'll go ahead and make that change as part of this PR. Though if you'd rather not have a change that large at this stage of the review, I'll hold off.

davidflanagan:
Things still to be done before landing this:

- [ ] John's request to add a commit that does a git mv lib/kumascript src at the start, to preserve the history of files like firelogger.js. I think I can do this with an `edit` step with `git rebase -i`. But I'm worried that it will mess up the comments and conversation, so I'm not doing it just yet.

-[ ] If Ryan agrees, move the old autorequire macro code back out of environment.js into files in src/api/. This will be neater, and will separate the untested code from the tested code. If I do this, I should also add a commit at the start that does a `git mv macros/DekiScript-*.ejs src/api/*.js`

-[ ] Integrate newrelic config information into config.js via environment variables and re-enable the newrelic initialization code in server.js

ExE-Boss:
The migration to&nbsp;Jest could be&nbsp;done in&nbsp;a&nbsp;separate&nbsp;PR, which&nbsp;would help with the&nbsp;<code>git&nbsp;mv</nbsp> issues.

I believe splitting this&nbsp;PR might make merging it&nbsp;easier. (also, reviewing it&nbsp;wouldn’t be&nbsp;such a&nbsp;pain on&nbsp;mobile)

jwhitlock:
> John's request to add a commit that does a git mv lib/kumascript src at the start, to preserve the history of files like firelogger.js. I think I can do this with an edit step with git rebase -i. But I'm worried that it will mess up the comments and conversation, so I'm not doing it just yet. (And actually, it looks like firelogger.js was correctly moved already. Maybe tests/.js and tests/macros/.js would benefit if I insert a commit that changes from the old mocha test-foo.js name to the new jest foo.test.js name.)

After sleeping on it, I'm willing to drop this. The consensus in issue #1034 was to move this code into the Kuma repo, but after some preliminary work. The most likely way to do that is import the code (macros plus server) in one commit, and pointing at this repo for history. Since we're going to have a KumaScript epoch in 2019 anyway, where there's a big break in history, let's not spend time adding the commit to preserve history for this repo.

macros/MakeSimpleQuicklinks.ejs
```diff
@@ -13,24 +13,24 @@
...
-var count = arguments.length;
+let quicklinks = "<section id='Quick_Links'><ol>";
+let count = arguments.length;
 
-for (var i=0; i<count; i++) {
-    var item = arguments[i];
-    var path = "";
-    var title = "";
-    var text = "";
+for (let i=0; i<count; i++) {
```
schalkneethling: Nit: Not introduced here but, the preferred code style is:

```suggestion
for (let i = 0; i < count; i++) {
```

src/cache.js
```diff
@@ -0,0 +1,70 @@
+/**
+ * This module defines a cache() function for caching the strings returned by
+ * other functions.
+ *
+ * KumaScript macros originally used memcached for caching. This
+ * module is a new backend to replace memcached. Currently it just use
+ * a LRU cache in local instance memory, which means that if we have
```
schalkneethling: Nit: "Currently it just *uses an* LRU cache"?

escattone:
@davidflanagan @jwhitlock I'm wondering if we should add `memcached` back into this (and replace it with `cache.js` or something else in a separate PR that we'll consider/merge/deploy later)? I think I'd prefer that slightly, but I'm also OK with just going with this as it stands (the local LRU cache in `cache.js`) if you guys are.

src/cache.js
davidflanagan: Fixed in the cache.js rewrite to use Redis.

macros/MakeSimpleQuicklinks.ejs
davidflanagan: Not going to touch that in this PR.

davidflanagan:
This is getting close to ready. Getting the newrelic configuration set up is the only remaining chunk of work that I know of.

src/cache-redis.js
```diff
@@ -0,0 +1,29 @@
...
+                    /* eslint-enable no-console */
+                    resolve(null);
+                } else {
+                    resolve(response);
+                }
+            });
+        });
+    },
+
+    set(key, value) {
```
escattone: I would vote for making this `async` as well, just like you did for the `get`. I'd prefer to have all I/O operations done asynchronously.

escattone:
@davidflanagan Wow, you modified the caching and its tests to handle redis so quickly and it's beautifully done as well, thanks! We should be able to finalize this tomorrow. I'm getting errors when I try to run the `make` commands which I don't yet understand, but that'll have to wait until tomorrow.

jwhitlock:
@davidflanagan I think you mentioned in another channel you needed some pointers for the New Relic config.

The staging and production configuration for KumaScript are set by [Jinja2 templates that generate yaml](https://github.com/mdn/infra/blob/8cf18dc1782bf8756ad3e50715628b3d14d5903f/apps/mdn/mdn-aws/k8s/kumascript.deploy.yaml.j2#L22-L67) in the mdn/infra repo. Some of the variation is applied in the [Makefile](https://github.com/mdn/infra/blob/8cf18dc1782bf8756ad3e50715628b3d14d5903f/apps/mdn/mdn-aws/k8s/Makefile#L130-L133), including the [New Relic configuration](https://github.com/mdn/infra/blob/8cf18dc1782bf8756ad3e50715628b3d14d5903f/apps/mdn/mdn-aws/k8s/Makefile#L197-L213), and is the same across AWS environments or can be derived from deployment-specific parameters. Deployment-specific parameters are set in script files such as [regions/oregon/prod.sh](https://github.com/mdn/infra/blob/8cf18dc1782bf8756ad3e50715628b3d14d5903f/apps/mdn/mdn-aws/k8s/regions/oregon/prod.sh#L135) or [regions/oregon/stage.sh](https://github.com/mdn/infra/blob/8cf18dc1782bf8756ad3e50715628b3d14d5903f/apps/mdn/mdn-aws/k8s/regions/oregon/stage.sh#L139). 

There may be differences between the development and AWS defaults for config that you've already migrated. The New Relic environment variables appear [toward the end of the list](https://github.com/mdn/infra/blob/8cf18dc1782bf8756ad3e50715628b3d14d5903f/apps/mdn/mdn-aws/k8s/kumascript.deploy.yaml.j2#L52-L67), and the license key is the only secret value. @escattone or I should be able to get you the "rendered" configuration from stage or production, minus the license key.

I seem to remember the [New Relic node.js](https://docs.newrelic.com/docs/agents/nodejs-agent/installation-configuration/nodejs-agent-configuration) documents being useful but confusing when we added New Relic to the AWS KumaScript [2 years ago](https://github.com/mozmeao/infra/pull/542). ``NEW_RELIC_MONITOR_MODE=0`` might have been useful locally to get New Relic running but not reporting.

davidflanagan:
Thanks Ryan! John: that is super-helpful info about the new relic config. Thank you.

escattone:
Thanks for adding more tests @davidflanagan and increasing our code coverage!
image

escattone:
On a final pass through, I realized that we need changes to the `Jenkinsfile`. They are not complicated, so I'm going to list them here:

- First, lines 5-21 (the `test` function) needs to change to this:
```groovy
def test(docker_tag='') {
    def jenkins_uid = sh(script: 'id -u jenkins', returnStdout: true).trim()
    // The user ID within the Docker container must match the "jenkins"
    // user ID so that it's able to create the "test-results.xml" file.
    def run_args = "-u ${jenkins_uid}" + ' -v ${PWD}:${APP_DIR} -w ${APP_DIR}'
    try {
        withEnv(["VERSION=${docker_tag}",
                 "DOCKER_RUN_ARGS=${run_args}",
                 'TEST_RUN_ARGS=--reporter mocha-junit-reporter']) {
            utils.sh_with_notify("make test",
                                 "Test the Kumascript code and macros")
        }
    } finally {
        junit 'test-results.xml'
    }
}
```

- Lines 23-29 (the `lint` function) needs to change to this:
```groovy
def lint(what, docker_tag='') {
    def make_what = (what == 'code') ? 'lint' : 'lint-json'
    withEnv(["VERSION=${docker_tag}"]) {
        utils.sh_with_notify("make ${make_what}",
                             "Lint the Kumascript ${what}")
    }
}
```

- Lines 31-36 (the `test_kumascript` function) needs to change to this:
```groovy
def test_kumascript(docker_tag='') {
    dir('kumascript') {
        test(docker_tag)
    }
}
```

- And finally, line 41 needs to change to this:
```groovy
        lint('json', docker_tag)
```

escattone:
Also, I think we should update the `README.md` to reflect your latest changes to the caching files, as well as perhaps add something about the `make test-coverage` command.

davidflanagan:
Thanks @escatone for catching the Jenkinsfile issue before the deploy! And for the reminder to update my documentation to reflect reality. (Please never let me slide on stale documentation comments!)

@jwhitlock: I've addressed Ryan's final review comments in https://github.com/mdn/kumascript/pull/1035/commits/e91bf4160df0229e7d16f52e3a583b9ee87bd6ea

jwhitlock:
Thanks @davidflanagan, and thanks to @escattone for the thorough review.

schalkneethling:
This is a HUGE milestone! Thanks everyone and @davidflanagan for all the codes!

PR #1037, opened 2019-01-11
Add macro snapshot tests

davidflanagan:
Most of our macros don't have explicit tests. This patch adds
a brittle alternative that attempts to render each untested macro
in a special hacky environment and compares its output to a
previously recorded "snapshot". This is very imperfect, but it
may catch accidental changes to macro .ejs files or to the macro
rendering code that would have otherwise gone unnoticed.

If intentional changes are made to a macro's output then the
snapshots need to be re-generated (by running
`node tests/macros/create-snapshots.js`). If this gets annoying
then you can write a real test for the macro, and regenerate the
snapshots one final time: macros that have tests are not included
in the snapshot test. Note that when you regenerate the snapshots
you should check the resulting tests/macros/snapshots.json file in.

The snapshot tests exclude macros that are .json files since there
is nothing to test there and exclude sidebar macros because those
actually change frequently for legitimate reasons. Two macros that
have randomized output are also excluded since they would always fail.

Finally, this patch also includes a bugfix for the CSS_Ref macro that
I found through snapshot testing.


wbamberg:
I'm worried that this will be annoying without being terribly helpful. If I understand correctly, it will fail every time (for example) someone adds a new item to a sidebar macro, and to fix it the person will have to copy in the new macro output. I'm worried that this will be an annoying extra step that people will just do automatically without thinking, so it won't catch any actual problems.

I've thought a bit about macro testing in the past. Obviously regression tests are useful when code is being changed. If we look at the history of changes to KS macros, there are 2 sorts of changes that make up most of them: (1) changes to sidebars and (2) changes to JSON data macros.

## Changes to sidebars

These updates are very common because sidebar macros are (partly, and variably) HTML content implemented in JS - so when [someone writes a new tutorial and want it to appear in the sidebar, they have to update the macro code](https://github.com/mdn/kumascript/pull/1051/files). It's most unfortunate, here, that the sidebar system doesn't distinguish data from code, so data updates require code changes.

The main risk here I think is people breaking the HTML output. For example: https://discourse.mozilla.org/t/why-is-all-normal-text-bold-now/31038 where an update to the content  didn't close a `strong` tag.

What I think would be helpful for this would be a test that validated the HTML output of sidebar macros. I started a thread on this here: https://discourse.mozilla.org/t/linting-html-output-of-ks-macros/31506 and @escattone noticed that the W3C HTML validator (https://validator.w3.org/#validate_by_input) has a Docker image, and suggested that we could use it in a docker-compose file to validate output.

## Changes to JSON data

We have various macros that are just JSON blobs - SpecData, GroupData, InterfaceData. These also get a lot of churn, because again they are data not code.

It's easy to make mistakes editing JSON, and this has been the cause, but also easy to test for them. So for these we should just write tests. I wrote a simple test for SpecData and am in the process of doing that for GroupData.

Ultimately we should consider moving this out of KS though, and into mdn/data, where we could validate the JSON using a schema. @ExE-Boss is working on that at https://github.com/mdn/kumascript/pull/1041 and if we want to do this we should support that work.

davidflanagan:
I've updated the PR to exclude data and sidebar macros.  The create-snapshots.js file also has an easy way to exclude other specific macros (see the doNotTest variable in the code) so there is an escape hatch if we start getting too many failures. Does this help @wbamberg ?

davidflanagan:
Looks like I've got lint to fix, and a snapshot test that passes locally and fails in Docker. I'll need to fix both of those if there is interest in landing this.

PR #1040, opened 2019-01-16
rm Embed_text.ejs and EmbedSVG.ejs

wbamberg:
These macros were used in 1 en-US and 50-odd pages across all locales. Most of the time they made you a helpful table like this:

screen shot 2019-01-16 at 1 46 45 pm

In these cases I just deleted the table. In a few cases the SVG code was actually there, so I just copied it into a static code block.

Now these macros are no longer called by any Wiki pages:
https://developer.mozilla.org/en-US/search?kumascript_macros=Embed_text&locale=*&page=1&topic=none
https://developer.mozilla.org/en-US/search?locale=*&kumascript_macros=EmbedSVG&topic=none

...and they're not called internally by any other macros:
https://github.com/mdn/kumascript/search?utf8=%E2%9C%93&q=Embed_text&type=
https://github.com/mdn/kumascript/search?utf8=%E2%9C%93&q=EmbedSVG&type=


jwhitlock:
@wbamberg there is a merge conflict because of the KumaScript refactor. Would you like to force-push, resubmit, or let someone else submit?

wbamberg:
@jwhitlock , I'll resubmit, thanks for the heads-up.

PR #1046, opened 2019-01-22
Add a Makefile target to run tests and output junit test results.

davidflanagan:
The Jenkinsfile expects a test report in junit format. This patch adds
a Jest plugin to generate that report and adds a Makefile target to
run tests using that plugin.


davidflanagan:
I added the --ci flag since this Makefile target is intended for use in a CI environment. That only affects snapshot test, and we don't have any, but it seemed like a good idea.

And I removed the -w1 flag. That was an optimization for running tests in my local Docker container, and may not be optimal for running tests on the Jenkins box. In this case I figure I should let Jest figure out how much parallelism to use.

PR #1047, opened 2019-01-22
rm EmbedSVG.ejs and Embed_text.ejs

wbamberg:
Resubmitting https://github.com/mdn/kumascript/pull/1040:

-------------

These macros were used in 1 en-US and 50-odd pages across all locales. Most of the time they made you a helpful table like this:

screen shot 2019-01-16 at 1 46 45 pm

In these cases I just deleted the table. In a few cases the SVG code was actually there, so I just copied it into a static code block.

Now these macros are no longer called by any Wiki pages:
https://developer.mozilla.org/en-US/search?kumascript_macros=Embed_text&locale=*&page=1&topic=none
https://developer.mozilla.org/en-US/search?locale=*&kumascript_macros=EmbedSVG&topic=none

...and they're not called internally by any other macros:
https://github.com/mdn/kumascript/search?utf8=%E2%9C%93&q=Embed_text&type=
https://github.com/mdn/kumascript/search?utf8=%E2%9C%93&q=EmbedSVG&type=


macros/Embed_text.ejs
```diff
@@ -1,39 +0,0 @@
...
-});
-
-if (brush.indexOf($1) > -1) {
-    syntaxe = $1;
-}
-
-for(let aFile of env.files) {
-    if ($0 === aFile.filename) {
-        found = true;
-%><pre class="brush: <%= syntaxe %>"><%= await MDN.getFileContent(aFile) %></pre><%
```
ExE-Boss: Fun fact: this would have broken anyway due to a logic error in `MDN.getFileContent(…)`: https://github.com/mdn/kumascript/blob/4807e01c869378841c66c8d3b96e2e5281c08ade/src/api/mdn.js#L127-L135

PR #1053, opened 2019-01-25
Bug 1522206: Define translation text for the WebExtAPISidebar macro.

davidflanagan:
The WebExtAPISidebar macro (used by AddonSidebar) tries to access
a text[] array that does not exist. This appears to be an old bug
that only recently started throwing an error after the KumaScript
refactor.

This PR takes the text array from the APIRef macro and uses a
tweaked version in WebExtAPISidebar. It also adds translations
to the sidebar strings "Properties", "Methods", "Types" and "Events".

### DIFF
--- a/macros/WebExtAPISidebar.ejs
+++ b/macros/WebExtAPISidebar.ejs
@@ -12,6 +12,17 @@ var escapeQuotes = mdn.escapeQuotes;
 var htmlEscape = kuma.htmlEscape;
 var rtlLocales = ['ar', 'he', 'fa'];
 
+let commonl10n = string.deserialize(await template('L10n:Common'));
+
+let text = {
+    'translate': mdn.getLocalString(commonl10n, '[Translate]'),
+    'title': mdn.getLocalString(commonl10n, 'TranslationCTA'),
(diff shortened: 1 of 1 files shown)


davidflanagan:
Before this PR:
screen shot 2019-01-25 at 11 28 55 am

After this PR:
screen shot 2019-01-25 at 11 29 53 am

jwhitlock:
@davidflanagan do you remember what doc you needed to reproduce the bug?

davidflanagan:
@jwhitlock: https://developer.mozilla.org/fr/docs/Mozilla/Add-ons/WebExtensions/API/bookmarks/create

I had to scrape both bookmarks and bookmarks/create and bookmarks/onCreated, and then find a locale (french in this case) where the onCreated event was not translated in order to trigger it.  As you can see from the screenshot, it also scraped the WebExtensions/API/i18n page, but I don't think that was necessary.

It was a 5 minute fix that took an hour of fumbling around with to figure out how to reproduce locally.

SphinxKnight:
Thanks :) Looks good to me too!

jwhitlock:
Thanks for the reviews @wbamberg and @SphinxKnight!

PR #1057, opened 2019-01-30
test: Fix tests on Windows

ExE-Boss:
The `templates` test was failing on Windows due to the usage of `\` in file paths, where the test expected `/`.

---

review?(@chrisdavidmills, @davidflanagan, @Elchi3, @jwhitlock, @peterbe, @wbamberg)


tests/templates.test.js
```diff
@@ -76,18 +77,18 @@ describe('Templates class', () => {
 
     it('only loads files once', async () => {
         const EJS = require('ejs');
-        const mockLoader = jest.fn(filename => `<%= "${filename}" -%>`);
+        const mockLoader = jest.fn(filename => `<%= ${JSON.stringify(filename)} -%>`);
```
ExE-Boss: Without this `\` in the file path would be treated as part of an escape sequence (e.g.: `C:\nodejs\...\kumascript\tests\fixtures\templates\test.ejs` would be interpreted as:
<pre><code>
C:
odejs...kumascript	ests&#x0C;ixtures	emplates	est.ejs
</code></pre>

tests/macros/utils.js
```diff
@@ -176,17 +176,22 @@ function lintHTML(html, fragment = true) {
                 </html>`;
     }
     try {
-        execSync(`java -jar ${vnu} --errors-only --format text -`, {
+        execSync(`java -jar ${JSON.stringify(vnu)} --errors-only --format text -`, {
```
ExE-Boss: Not&nbsp;using `JSON.stringify(…)` here results in&nbsp;a&nbsp;problem similar to&nbsp;https://github.com/mdn/kumascript/pull/1057#discussion_r256427946, but&nbsp;with spaces instead.

tests/macros/utils.js
```diff
@@ -176,17 +176,22 @@ function lintHTML(html, fragment = true) {
...
             stdio: 'pipe',
             timeout: 15000
         });
         return null;
     } catch (error) {
         const error_message = error.message
-            .split(os.EOL)
-            .filter(line => line.startsWith('Error: '))
+            // `vnu` always uses `\n`, even on Windows.
+            .split(/\r?\n/g)
```
peterbe: If it's always `\n` (even on Windows) then why split on possibly `\r`?

tests/macros/utils.js
peterbe: That's a very useful comment, in https://github.com/mdn/kumascript/pull/1057#discussion_r256427946
Can you transpose that into a code comment right above line 179?

tests/templates.test.js
peterbe: Same here. That's a very useful comment. Let's play Jeopardy, the comment should be the answer to "Why does the file name need to be JSON.stringify escaped on this line???"

tests/macros/utils.js
ExE-Boss: In a future case if `vnu` ever gets updated and starts adding `\r` on Windows.

PR #1065, opened 2019-02-03
Add tests for GroupData

wbamberg:
This adds a test for GroupData.json. There's no spec for this structure, so I've looked at the macros that use it to try to figure out what kinds of assumptions they make about the data.

I've treated four of the properties - `interfaces`, `methods`, `properties`, and `events` - as mandatory arrays even if they are empty. As far as I can tell only `events` is actually assumed to be present (in [eventref.ejs](https://github.com/mdn/kumascript/blob/master/macros/eventref.ejs#L36)). However, since `properties` is already present in all groups, and `methods` and `interfaces` are present in all but one group ([and one of those is apparently an error](https://github.com/mdn/kumascript/blob/master/macros/GroupData.json#L1438)), well, it seemed simpler and safer to require all four in the tests, and to add the missing ones to the data.

I've tried to require that strings contain only a restricted set of characters, based on what they already contain and what seems sensible. Macros that use this data sometimes [insert it into the HTML output](https://github.com/mdn/kumascript/blob/master/macros/APIRef.ejs#L263) without any sanitation, so for example stuff like this:

```
        "Alarm API": {
            "overview":   [ "Alarm<strong> API"],
            "interfaces": [ "MozAlarmsManager" ],
            "methods":    [ "Navigator.mozSetMessageHandler()"],
            "properties": [ "Navigator.mozAlarms" ],
            "events":     []
        },
```

...would break pages.

I've tried to use the new testing framework to some extent anyway. I've used `expect` instead of `assert`, but I've not restructured the code to enable (the equivalent of) assertion messages by making every assertion a separate test, since that seems like that will make this code more complicated and verbose. I guess if we decide to use https://github.com/mattphillips/jest-expect-message then it's easy to add assertion messages to this code.

### DIFF
--- a/macros/GroupData.json
+++ b/macros/GroupData.json
@@ -1435,7 +1435,7 @@
             "events":     [ "change" ]
         },
         "Selection API": {
-            "interface":  [ "Selection" ],
+            "interfaces":  [ "Selection" ],
             "properties": [ "Document.onselectionchange",
                             "GlobalEventHandlers.onselectstart" ],
             "methods":    [ "Document.getSelection()",
@@ -1893,6 +1893,7 @@
                             "AuthenticatorResponse",
                             "AuthenticatorAttestationResponse",
                             "AuthenticatorAssertionResponse" ],
+            "methods":    [],
             "properties": [],
             "events":     []
         },
--- a/macros/GroupData.json
+++ b/macros/GroupData.json
@@ -1435,7 +1435,7 @@
             "events":     [ "change" ]
         },
         "Selection API": {
-            "interfaces":  [ "Selection" ],
+            "interfaces": [ "Selection" ],
             "properties": [ "Document.onselectionchange",
                             "GlobalEventHandlers.onselectstart" ],
             "methods":    [ "Document.getSelection()",
--- a/macros/GroupData.json
+++ b/macros/GroupData.json
@@ -658,7 +658,7 @@
                             "gamepaddisconnected" ]
         },
         "Geolocation API": {
-            "guide":      [ { "url":   "/en-US/docs/Web/API/Geolocation/Using_geolocation",
+            "guides":     [ { "url":  "/en-US/docs/Web/API/Geolocation/Using_geolocation",
                               "title": "Using geolocation" } ],
             "interfaces": [ "Coordinates",
                             "Geolocation",
--- a/macros/GroupData.json
+++ b/macros/GroupData.json
@@ -658,7 +658,7 @@
                             "gamepaddisconnected" ]
         },
         "Geolocation API": {
-            "guides":     [ { "url":  "/en-US/docs/Web/API/Geolocation/Using_geolocation",
+            "guides":     [ { "url":   "/en-US/docs/Web/API/Geolocation/Using_geolocation",
                               "title": "Using geolocation" } ],
             "interfaces": [ "Coordinates",
                             "Geolocation",
--- a/tests/macros/groupdata.test.js
+++ b/tests/macros/groupdata.test.js
@@ -52,7 +52,7 @@ function checkGroupData(groupDataJson) {
   for (let groupName of groupNames) {
     // the group name contains only the permitted characters
     expect(permittedCharacters.group.test(groupName)).toBe(true);
-    group = groupData[0][groupName];
+    const group = groupData[0][groupName];
 
     // the group contains interfaces, properties, methods and events
     // they are all string arrays and contain only the permitted characters
--- a/tests/macros/groupdata.test.js
+++ b/tests/macros/groupdata.test.js
@@ -15,22 +15,77 @@ const permittedCharacters = {
   property:   /^[\w.]+$/,
   method:     /^[\w.()]+$/,
   event:      /^[\w()]+$/,
+  dictionary: /^\w+$/,
+  callback:   /^\w+$/,
+  type:       /^\w+$/,
   guideTitle: /^[\w .,]+$/,
   guideUrl:   /^\/[\w-.~/]+$/
 }
 
 /*
-Given `strings`, which should be an array of strings,
-and `characters`, which is a regular expression defining
-the permitted characters in each string, this checks
-that `strings` really is an array and that the strings
-it contains contain only the permitted characters.
+Properties that are allowed in a group
 */
-function checkStringArray(strings, characters) {
-  expect(strings).toBeDefined();
+const permittedGroupProperties = [
+  'overview',
+  'interfaces',
+  'methods',
+  'properties',
+  'dictionaries',
+  'callbacks',
+  'types',
+  'events',
+  'guides'
+];
+
+/*
+Properties that must be present in a group
+*/
+const mandatoryGroupProperties = [
+  'interfaces',
+  'methods',
+  'properties',
+  'events',
+];
+
+/*
+Properties that are allowed in a guide
+*/
+const permittedGuideProperties = [
+  'title',
+  'url'
+];
+
+/*
+Properties that must be present in a guide
+*/
+const mandatoryGuideProperties = [
+  'title',
+  'url'
+];
+
+/*
+Check that `obj` contains:
+* only the properties in `permitted`
+* all the properties in `mandatory`
+*/
+function checkProperties(obj, permitted, mandatory) {
+  let props = Object.keys(obj);
+  for (let prop of props) {
+    expect(permitted.includes(prop)).toBe(true);
+  }
+  for (let prop of mandatory) {
+    expect(props.includes(prop)).toBe(true);
+  }
+}
+
+/*
+Check that `strings` is an array of strings,
+and that each string matches the given regex.
+*/
+function checkStringArray(strings, permitted) {
   expect(Array.isArray(strings)).toBe(true);
   for (let string of strings) {
-    expect(characters.test(string)).toBe(true);
+    expect(permitted.test(string)).toBe(true);
   }
 }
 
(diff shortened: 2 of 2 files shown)


wbamberg:
@a2sheppy , @chrisdavidmills please let me know if the tests here match your expectation for what the structure of GroupData ought to be.

macros/GroupData.json
```diff
@@ -1893,6 +1893,7 @@
                             "AuthenticatorResponse",
                             "AuthenticatorAttestationResponse",
                             "AuthenticatorAssertionResponse" ],
+            "methods":    [],
             "properties": [],
             "events":     []
```
ExE-Boss: These empty arrays are actually unnecessary.

ExE-Boss:
See&nbsp;also&nbsp;https://github.com/mdn/data/pull/342, which shows that there are even more errors in&nbsp;the&nbsp;data.

macros/GroupData.json
wbamberg: Please see my PR description where I talk about these.

wbamberg:
> See also mdn/data#342, which shows that there are even more errors in the data.

Thanks! As far as I can tell the only extra error there is [the one for the Geolocation API](https://github.com/mdn/data/pull/342/files#r246139024), right? I just fixed that one.

I also added tests to check there aren't extra properties, to catch things like this, and in the process found some I wasn't testing: `dictionaries`, `callbacks` and `types`.

chrisdavidmills:
Looks OK to me from a brief glance, but I'm really snowed under this week.

@a2sheppy , you looked at this stuff in detail fairly recently. Can you give this a proper review please?

macros/GroupData.json
a2sheppy: There is some code out there that fails if these aren't included. I learned that the hard way a couple months ago.

a2sheppy:
> > See also [mdn/data#342](https://github.com/mdn/data/pull/342), which shows that there are even more errors in the data.
> 
> Thanks! As far as I can tell the only extra error there is [the one for the Geolocation API](https://github.com/mdn/data/pull/342/files#r246139024), right? I just fixed that one.
> 
> I also added tests to check there aren't extra properties, to catch things like this, and in the process found some I wasn't testing: `dictionaries`, `callbacks` and `types`.

Yes, these are new; we will be moving stuff that was previously jammed into `interfaces` into these over time. Sidebar code will need to be updated to know to look for them. Same for macros that build lists of interface members.

tests/macros/groupdata.test.js
```diff
@@ -0,0 +1,161 @@
...
+const { itMacro, describeMacro } = require('./utils');
+
+/*
+Different strings in GroupData objects have different sets of
+permitted characters.
+*/
+const permittedCharacters = {
+  group:      /^[\w ()-]+$/,
+  overview:   /^[\w ()-]+$/,
+  interface:  /^[\w.]+$/,
```
a2sheppy: Is it worth testing for the first character of an interface name being upper-case, per the rules for naming interfaces?

tests/macros/groupdata.test.js
```diff
@@ -0,0 +1,161 @@
...
+  property:   /^[\w.]+$/,
```
a2sheppy: Similarly, property and method names always start with a lower-case letter (except for constructors).

tests/macros/groupdata.test.js
```diff
@@ -0,0 +1,161 @@
...
+  method:     /^[\w.()]+$/,
+  event:      /^[\w()]+$/,
+  dictionary: /^\w+$/,
+  callback:   /^\w+$/,
+  type:       /^\w+$/,
+  guideTitle: /^[\w .,]+$/,
```
a2sheppy: Does this allow for colons? Doesn't look like it, but I know we have titles that include them.

tests/macros/groupdata.test.js
wbamberg: Thanks @a2sheppy , I've made this change.

tests/macros/groupdata.test.js
wbamberg: There are many cases where this isn't true in the data, e.g. https://github.com/mdn/kumascript/blob/master/macros/GroupData.json#L6.

We could make the validation a bunch more complex here (split on dots, and validate the pieces separately) but I don't think it's worth it. The main point of this test is to make sure that assumptions made by the macros aren't broken by data updates.

tests/macros/groupdata.test.js
wbamberg: It does not, but there aren't any guides in here with colons (or the test would fail). I've chosen to be conservative with things like this: rather than permit any characters that it seems we might want to use, I'm permitting the ones we actually do use. If someone adds a guide that contains a colon, they can update the regex then.

tests/macros/groupdata.test.js
```diff
@@ -0,0 +1,161 @@
...
+      checkStringArray(group.callbacks, permittedCharacters.callback);
+    }
+    if (group.types) {
+      checkStringArray(group.types, permittedCharacters.type);
+    }
+
+    // overview is optional
+    if (group.overview) {
+      // if present it must contain only the permitted characters
+      expect(permittedCharacters.overview.test(group.overview)).toBe(true);
```
ExE-Boss:
```suggestion
      expect(group.overview).toMatch(permittedCharacters.overview);
```

tests/macros/groupdata.test.js
```diff
@@ -0,0 +1,161 @@
...
+}
+
+/*
+Check that `strings` is an array of strings,
+and that each string matches the given regex.
+*/
+function checkStringArray(strings, permitted) {
+  expect(Array.isArray(strings)).toBe(true);
+  for (let string of strings) {
+    expect(permitted.test(string)).toBe(true);
```
ExE-Boss:
```suggestion
    expect(string).toMatch(permitted)
```

tests/macros/groupdata.test.js
```diff
@@ -0,0 +1,161 @@
...
+
+/*
+Check that `obj` contains:
+* only the properties in `permitted`
+* all the properties in `mandatory`
+*/
+function checkProperties(obj, permitted, mandatory) {
+  let props = Object.keys(obj);
+  for (let prop of props) {
+    expect(permitted.includes(prop)).toBe(true);
```
ExE-Boss:
```suggestion
    expect(permitted).toContain(prop);
```

tests/macros/groupdata.test.js
```diff
@@ -0,0 +1,161 @@
...
+  }
+  for (let prop of mandatory) {
+    expect(props.includes(prop)).toBe(true);
```
ExE-Boss:
```suggestion
    expect(props).toContain(prop);
```

tests/macros/groupdata.test.js
```diff
@@ -0,0 +1,161 @@
...
+  expect(groupData.length).toBe(1);
+
+  // the one object has one property for each group
+  // the property's key is the group name
+  const groupNames = Object.keys(groupData[0]);
+
+  for (let groupName of groupNames) {
+
+    // the group name contains only the permitted characters
+    expect(permittedCharacters.group.test(groupName)).toBe(true);
```
ExE-Boss:
```suggestion
    expect(groupName).toMatch(permittedCharacters.group);
```

tests/macros/groupdata.test.js
```diff
@@ -0,0 +1,161 @@
...
+    // guides is optional
+    if (group.guides) {
+      // if present it is an array of guides
+      expect(Array.isArray(group.guides)).toBe(true);
+      for (let guide of group.guides) {
+        // check that the guide has the correct properties
+        checkProperties(guide, permittedGuideProperties, mandatoryGuideProperties);
+
+        // these are both strings that contain only the permitted characters
+        expect(permittedCharacters.guideTitle.test(guide.title)).toBe(true);
```
ExE-Boss:
```suggestion
        expect(guide.title).toMatch(permittedCharacters.guideTitle);
```

tests/macros/groupdata.test.js
```diff
@@ -0,0 +1,161 @@
...
+        expect(permittedCharacters.guideUrl.test(guide.url)).toBe(true);
```
ExE-Boss:
```suggestion
        expect(guide.url).toMatch(permittedCharacters.guideUrl);
```

tests/macros/groupdata.test.js
a2sheppy: Fair enough. I was thinking we could protect against certain typos, too, but you're right that that's outside the scope of verifying the structure of the data.

tests/macros/groupdata.test.js
a2sheppy: Makes sense. I do think we need to be sure to note somewhere that we keep track of that the lack of permitting characters in this test isn't meant to suggest a rule about what's allowed in titles and such then.

a2sheppy:
As far as I'm concerned, once the changes suggested by @ExE-Boss are in, this is good to go. Nice work! 👍

wbamberg:
> To address @escattone’s concerns regarding error messages, I suggest using the proper Jest Matchers methods:

This doesn't help much with that problem, but it is much more readable code, so thanks!

> Also, I’d suggest that the file gets renamed to GroupData.test.js to match the capitalisation of GroupData.json.

I haven't done this. There are no standards for this AFAIK, and to the small extent that there is a convention, lowercase seems to be winning.

> Also, please convert the function comments to JSDoc comments.

Done!

ExE-Boss:
> This doesn't help much with that problem, but it is much more readable code, so thanks!

It does however output the&nbsp;exact string that is&nbsp;causing the&nbsp;test to&nbsp;fail and the&nbsp;`RegExp` that should match, so&nbsp;you&nbsp;can do&nbsp;a&nbsp;<kbd>Ctrl&nbsp;+&nbsp;F</kbd> and&nbsp;fix the&nbsp;issue.

PR #1068, opened 2019-02-04
Add new docs to Web Crypto API GroupData

wbamberg:
This adds `CryptoKeyPair` and all the `*Params` dictionaries to GroupData, in response to @chrisdavidmills 's comment: https://github.com/mdn/sprints/issues/664#issuecomment-458269382.

I'm not quite sure if I am doing this right. I've tested this locally: the effect is not fully visible in the sidebar because the APIRef sidebar macro doesn't yet understand about dictionaries.


a2sheppy:
Before I even review: Yes, you did the dictionaries right, other than that they should be alphabetical to help us quickly find things while editing, and to help avoid duplicates creeping in.

macros/GroupData.json
```diff
@@ -1932,14 +1932,31 @@
...
+            "overview":     [ "Web Crypto API" ],
+            "interfaces":   [ "Crypto",
+                              "CryptoKey",
+                              "CryptoKeyPair",
+                              "RandomSource",
+                              "SubtleCrypto" ],
+            "methods":      [],
+            "properties":   [ "Window.crypto" ],
+            "events":       [],
+            "dictionaries": [ "RsaPssParams",
```
a2sheppy: The dictionaries should be listed alphabetically.

wbamberg: Done! Note though that in most cases in this file "dictionaries" is not sorted. We should lint for this kind of thing.

wbamberg:
Thanks for the review @a2sheppy !

PR #691, opened 2018-04-19
Fix InterfaceOverview to properly close `<dl>`s

a2sheppy:
Also removed the Experimental badge and merged the
Obsolete badge with the Deprecated one.

PR #1073, opened 2019-02-07
Fix InterfaceOverview to properly close `<dl>`s

ExE-Boss:
Supersedes and closes #691

## TODO:
- [x] Write&nbsp;tests
- [x] Address&nbsp;https://github.com/mdn/sprints/issues/906#issuecomment-461314733

---

review?(@a2sheppy, @chrisdavidmills, @Elchi3, @wbamberg)

Depends on: https://github.com/mdn/kumascript/pull/1093


ExE-Boss:
**Travis&nbsp;CI** is&nbsp;once again not running the&nbsp;build. 🤦🏻‍♂️

tests/macros/InterfaceOverview.test.js
```diff
@@ -0,0 +1,80 @@
...
+ * @prettier
+ */
+
+// Get necessary modules
+const fs = require('fs');
+const path = require('path');
+
+const { assert, describeMacro, beforeEachMacro, itMacro } = require('./utils');
+const jsdom = require('jsdom');
+const { JSDOM } = jsdom;
```
davidflanagan: It would be simple to combine these two lines into one, since you don't use jsdom on its own anywhere.

tests/macros/InterfaceOverview.test.js
```diff
@@ -0,0 +1,80 @@
...
+
+const INTERFACE_DATA = `[${JSON.stringify({
+    XMLHttpRequestEventTarget: {
+        inh: 'EventTarget',
+        impl: []
+    }
+})}]`;
```
davidflanagan: What's your reasoning for using JSON.stringify() here instead of just writing the string? Is that so you get syntax checking? I'm intrigued by that idea. I think that for something as short as this, I'd just write it as as string rather than bothering with  the two step thing.

If you want to keep it this way, please at least move the square brackets inside the stringify call.

tests/macros/InterfaceOverview.test.js
```diff
@@ -0,0 +1,80 @@
...
+    /** @type {Record<string, object>} */
+    let result = {};
+    for (let i = 0; i < dataNames.length; i++) {
+        const name = dataNames[i];
+        result[name] = JSON.parse(
+            fs.readFileSync(path.resolve(DATA_DIR, `${name}.json`))
+        );
+    }
+    return result;
+})('EventTarget', 'XMLHttpRequestEventTarget');
```
davidflanagan: This would be clearer if you had a named function called readFixture, and then you could write

const FIXTURES = { EventTarget: readFixture('interface-overview', 'EventTarget'), ...};

I see that other fixtures already exist. Which tests use them and how do those tests read them in? How about adding this readFixture() function to utils.js?

tests/macros/InterfaceOverview.test.js
```diff
@@ -0,0 +1,80 @@
...
+
+/**
+ * @param {string[]} headers
+ * @returns {(result: string) => void}
+ */
+function testInterface(headers) {
```
davidflanagan: A name like checkSectionTitles() would make the code clearer.

tests/macros/InterfaceOverview.test.js
```diff
@@ -0,0 +1,80 @@
...
+    return result => {
```
davidflanagan: I don't like that this is a function that returns a function because down below where you use it, it looks like a code error.

I'd prefer to see code below that was like `.then(result => checkSectionTitles(result, [...])`

I think that would make it a lot easier to understand.

tests/macros/InterfaceOverview.test.js
```diff
@@ -0,0 +1,80 @@
...
+    beforeEachMacro(macro => {
+        const realTemplate = macro.ctx.template;
+        macro.ctx.template = jest.fn(async (name, args) => {
+            switch (String(name).toLowerCase()) {
+                case 'interfacedata':
+                    return INTERFACE_DATA;
+                default:
+                    return realTemplate(name, args);
+            }
+        });
```
davidflanagan: This is a really interesting thing you're doing here... I think this is also worth putting in utils.js, with a name like mockMacro() or defineMacroMock().  I think that it would also need a corresponding unmockMacro() or removeMacroMock() function as well.  

(That doesn't have to happen in this PR if you don't want, however.)

tests/macros/InterfaceOverview.test.js
```diff
@@ -0,0 +1,80 @@
...
+        macro.ctx.page.subpagesExpand = jest.fn(async (path, depth, self) => {
+            let match = path && /\b[^/]+$/.exec(path);
+            return match && DATA[match[0]];
```
davidflanagan: It occurs to me that you could just read the fixture file here rather than loading it in advance above.  (If there was a readFixture() utility function)

tests/macros/InterfaceOverview.test.js
```diff
@@ -0,0 +1,80 @@
...
+        });
+    });
+
+    itMacro('No content in preview', async macro => {
```
davidflanagan: The title of this test confuses me. Would it be better as "No content without slug"?  Or maybe just as JavaScript comment explaining what you're testing here.

tests/macros/InterfaceOverview.test.js
```diff
@@ -0,0 +1,80 @@
...
+
+    itMacro('Interface with Constructor (EventTarget)', async macro => {
+        macro.ctx.env.slug = 'Web/API/EventTarget';
+        return macro.call().then(testInterface(['Constructor', 'Properties', 'Methods']));
+    });
+
+    itMacro('Interface without Constructor (XMLHttpRequestEventTarget)', async macro => {
+        macro.ctx.env.slug = 'Web/API/XMLHttpRequestEventTarget';
+        return macro.call().then(testInterface(['Properties', 'Methods']));
+    });
```
davidflanagan: Given that your bug fix is ensuring that <dl> tags are properly closed, do you need a test to ensure that the returned HTML is well-formed?  Or is JSDOM already validating that?  (If so, that is worth mentioning in a comment somewhere).

tests/macros/fixtures/interface-overview/EventTarget.json
```diff
@@ -0,0 +1,467 @@
+[
```
davidflanagan: An alternative to these long fixture files would be to cut them way down to the minimum and just code them directly into the test file. But if you think that other tests might be able to use them as well, that is a good reason to use fixture files.

davidflanagan (changes requested):
Thanks for this @ExE-Boss. I hate to say this, but could you submit a different version where you *don't* run Prettier on the macro? The formatting changes make it harder to figure out what the bug fix is.

Also, I don't see anything in this PR or the one it supersedes that explains what the bug is that is being fixed. I need a bit of context so I understand what is being fixed before I review the fix.

I have reviewed your test file and have left some comments on that. I've got some style nits there, and some suggestions for code that could usefully be moved into utils.js for use by other tests.

tests/macros/InterfaceOverview.test.js
ExE-Boss: I’m doing this in quite a few of my PRs (#1069 comes to mind: https://github.com/mdn/kumascript/blob/4734ac69c79855b0c40abedda3a99b1b82044428/tests/macros/DefaultAPISidebar.test.js#L35-L45

tests/macros/fixtures/interface-overview/EventTarget.json
ExE-Boss: The only thing I stripped from these files are non en‑US locales (e.g.:&nbsp;the&nbsp;content&nbsp;of&nbsp;`translations`).

tests/macros/InterfaceOverview.test.js
ExE-Boss: I’ve now added this to&nbsp;a&nbsp;comment: https://github.com/mdn/kumascript/blob/e0042e0c4c685c275e924c64d77c75eb751c5ad4/tests/macros/InterfaceOverview.test.js#L39-L46

macros/InterfaceOverview.ejs
```diff
@@ -36,28 +36,23 @@ if (slug) {
...
-        var webAPIGroups = string.deserialize(await template('GroupData'));
-    }
-
     var commonl10n = string.deserialize(await template('L10n:Common'));
 
     // Strings that need translating
 
     var strNone = localString({
-      "en-US": "This interface has no $1."
+        'en-US': 'This interface has no $1.'
```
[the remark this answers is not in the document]
ExE-Boss: Technically, it requires commenting out all the&nbsp;EJS&nbsp;`<%`,&nbsp;`%>` and&nbsp;HTML&nbsp;tags and then using the&nbsp;`‑‑parser typescript` command‑line option

tests/macros/InterfaceOverview.test.js
ExE-Boss: I prefer loading the data in&nbsp;advance (the&nbsp;`{{Compat}}` macro&nbsp;test does that too).

tests/macros/InterfaceOverview.test.js
```diff
@@ -0,0 +1,87 @@
...
+]);
+
+const DATA = ((...dataNames) => {
+    /** @type {Record<string, any>} */
+    let result = {};
+    for (const name of dataNames) {
+        result[name] = readJSONFixture('interface-overview', name);
+    }
+    return result;
+})('EventTarget', 'XMLHttpRequestEventTarget');
```
davidflanagan: I still think that this code is clever instead of being clear and straightforward. Also DATA is a really undescriptive name. I'd suggest:

```
const FIXTURES = {
    EventTarget: readJSONFixture('interface-overview', 'EventTarget'),
    XMLHttpRequestEventTarget: readJSONFixture('interface-overview', 'XMLHttpRequestEventTarget')
};
```

That is shorter and easier to understand than in immediately-invoked anonymous function. Remember that this code base is largely created and maintained by technical writers, not software engineers, so I put a real premium on making the code straightforward.

tests/macros/InterfaceOverview.test.js
```diff
@@ -0,0 +1,87 @@
...
+
+/**
+ * @param {DocumentFragment} dom
+ * @param {string[]} headers
+ */
+function checkSectionTitles(dom, headers) {
+    // An unclosed <dl> would just wrap all the remaining <h2>s,
+    // which would no longer be directly accessible through
+    // the `dom.children` property, as they’re no longer
+    // direct descendants.
```
davidflanagan: I know I asked a comment about whether these tests were actually testing the dl bug fix. So thank you for explaining this to me. But this comment here won't make any sense to any future reader who has not read this conversation on github, so please remove these lines.  

Instead, you could add something to the doc comment that just says that it verifies that dom element has h2 elements as direct children with content (and in the same order) as the headers array.

tests/macros/InterfaceOverview.test.js
```diff
@@ -0,0 +1,87 @@
...
+        }
+    );
+
+    itMacro('Unknown interface (Unknown)', async macro => {
+        macro.ctx.env.slug = 'Web/API/Unknown';
+        return macro.call().then(result => {
+            const dom = JSDOM.fragment(result);
+            checkSectionTitles(dom, ['Properties', 'Methods']);
+        });
+    });
```
davidflanagan: Looks like none of the tests excercise the code that outputs event handlers. If you don't have fixture data that would support that, I don't feel like you have to add it now. But it would be worth adding a comment indicating that that part of the macro is untested.

tests/macros/utils.js
```diff
@@ -63,15 +65,37 @@ function createMacroTestObject(macroName) {
...
+    let ctx = environment.prototypeEnvironment;
+
+    /** @type {Map<string, string>} */
+    const macroResults = new Map();
+    const realTemplate = ctx.template;
+    ctx.template = jest.fn(async (name, ...args) => {
+        let macroName = String(name)
+            .replace(/:/g, '-')
+            .toLowerCase();
+        return macroResults.get(macroName) || realTemplate(name, ...args);
```
davidflanagan: I think you should have an `await` before realTemplate. Otherwise, one case returns a string and the other case returns a Promise<string>. It may work, but would probably make a type checker angry.

davidflanagan: Also: do you think it will ever be important to be able to mock a template so that it returns an empty string? That won't be possible just using || here. How about if the get() call returns null or undefined then you call realTemplate(), but the empty string is a legal value?

tests/macros/utils.js
```diff
@@ -63,15 +65,37 @@ function createMacroTestObject(macroName) {
...
          * For example, "macro.ctx.env.locale" can be manipulated to something
          * other than 'en-US' or "macro.ctx.wiki.getPage" can be mocked
-         * using "sinon.stub()" to avoid network calls.
+         * using `jest.fn()` to avoid network calls.
+         */
+        ctx,
+
+        /**
+         * A much easier way to handle overwriting the result of the `template`
+         * function for specific macro calls.
```
davidflanagan: Could you expand on this comment, please? Something like:

"When writing tests for a macro that invokes other macros with the template() function
you sometimes want to specify a mock return value for those other macros. This function..."

And also in your comment you might mention that you can turn off a mock by calling the function again and passing null as the result.

tests/macros/utils.js
```diff
@@ -63,15 +65,37 @@ function createMacroTestObject(macroName) {
...
+         *
+         * @param {string} name
+         * @param {string} result
          */
-        ctx: environment.prototypeEnvironment,
+        mockTemplate(name, result) {
+            macroResults.set(name.replace(/:/g, '-').toLowerCase(), result);
```
davidflanagan: Optionally, you could add code here so that if result is null, then instead of setting that in the Map, you delete the name from the map instead.

macros/InterfaceOverview.ejs
```diff
@@ -36,11 +36,6 @@ if (slug) {
     // Fetch the databases
 
     var webAPIData = string.deserialize(await template('InterfaceData'));
-
-    if (group) {
-        var webAPIGroups = string.deserialize(await template('GroupData'));
-    }
```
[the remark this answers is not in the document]
davidflanagan: The easiest thing to do might be to leave all of this group-related code in the macro for now and just focus on the dl fix and the utils.js improvements. I won't be able to sign off on a patch that removes the group stuff without input from the writers who wrote and use the macro.

tests/macros/InterfaceOverview.test.js
```diff
@@ -0,0 +1,87 @@
...
+    // which would no longer be directly accessible through
+    // the `dom.children` property, as they’re no longer
+    // direct descendants.
+    let children = Array.from(dom.children).filter(n => n.matches('h2'));
+    expect(children.map(h => h.textContent.trim())).toEqual(headers);
+}
+
+describeMacro('InterfaceOverview', () => {
+    beforeEachMacro(macro => {
+        macro.mockTemplate('GroupData', INTERFACE_DATA);
```
davidflanagan: If you do end up deleting the group-related code in the macro, then you should also delete this line and the definition of INTERFACE_DATA above, I think.

tests/macros/utils.js
```diff
@@ -4,6 +4,8 @@
 
 // Provides utilities that as a whole constitute the macro test framework.
 
+const fs = require('fs');
+const path = require('path');
```
davidflanagan: nit: I like putting a blank line between different import sections. Import builtin modules, then a blank line, then import any libraries, then a blank line, then local imports.

tests/macros/utils.js
```diff
@@ -154,11 +178,31 @@ function afterEachMacro(teardown) {
     });
 }
 
+/**
+ * Reads a JSON fixture file from the `fixtures` directory.
+ *
+ * @param {...string} filePath
+ */
+function readJSONFixture(...filePath) {
```
davidflanagan: Do you think that fixtures will always be JSON-formatted and that tests will always want them in parsed form? What if you wanted to use a fixture file for a string that you were going to pass to mockTemplate()?

I think it would be more versatile to rename this to readFixture() and have it return a string. Then individual tests can decide whether to call parseJSON() or not. Or I suppose you could just define both readFixture() and readJSONFixture()

tests/macros/utils.js
```diff
@@ -154,11 +178,31 @@ function afterEachMacro(teardown) {
...
+    let fileName = filePath.pop();
+    if (!path.extname(fileName)) {
+        fileName += '.json';
```
davidflanagan: This is a nice touch!

tests/macros/utils.test.js
```diff
@@ -0,0 +1,11 @@
+/**
+ * @prettier
```
davidflanagan: Thank you for adding this test.

Since you are adding the @prettier comments to your tests, did you remember to `npm run format` before submitting the PR? (I don't see anything that looks incorrectly formatted... just asking because I forget to do it myself.)

davidflanagan (changes requested):
@ExE-Boss: I'd like another round of changes. In particular I don't know what the right thing to do about the groupdata changes. I think we'd need Chris or @a2sheppy to weigh in on that, so it might be simpler to just not change the group stuff at all in this PR.

I've got a few suggestions about the changes to utils.js, but I think they're going to be really useful for other tests. @escattone and @wbamberg: I've suggested that ExE-Boss add readJSONFixture() and macro.mockTemplate() to the test utilities. If you have a chance, please take a look at those new functions and how they're used here in the InterfaceOverview.test.js file.

tests/macros/utils.js
ExE-Boss: Since the function is `async`, returning a promise or non&#x2011;promise is the same as returning a promise in both cases or using `await` (Promises are auto&#x2011;flattened in&nbsp;JS)

tests/macros/utils.test.js
ExE-Boss: I run <code>[pnpx](https://pnpm.js.org) prettier <path&#x2011;to&#x2011;file> &#x2011;&#x2011;write</code> manually.

tests/macros/utils.js
ExE-Boss: Also, I’ll handle that case too.

(I keep forgetting that `!!''` results in&nbsp;`false` in JavaScript)

tests/macros/InterfaceOverview.test.js
ExE-Boss: I accidentally wrote `GroupData` instead of `InterfaceData`.

tests/macros/utils.js
davidflanagan: I understand that you *can* return a promise or non-promise from an async function. But I think it is bad style to use promise and non-promise on a single line like this, and I suspect that a type checker like Flow would not allow it. Please add the await.

tests/macros/utils.test.js
davidflanagan: If you use the 'format' script in package.json, it will ensure that only files with the @prettier comment get formatted. It doesn't matter how you run it, though, as long as you're using whatever version of prettier is installed in package.json, and not a global install.

macros/InterfaceOverview.ejs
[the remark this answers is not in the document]
a2sheppy: So, I created `InterfaceOverview` originally and generally am the one who has been working on it. I have plans to make further improvements and add more features but have been stalled because of the problems I've had trying to get working tests in place. Once tests are in, I can get back to work on this.

Ideally, this PR will simply fix the missing `</dl>` and add the needed tests. I can then follow up by making the not insignificant code changes I have in mind.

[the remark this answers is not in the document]
ExE-Boss: I’ve removed the currently unused dependency on `GroupData` to unblock work on https://github.com/mdn/sprints/issues/906, which was requested by @chrisdavidmills.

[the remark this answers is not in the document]
davidflanagan: I propose that you leave the group-related lines of code in, but comment them out. Since they currently have no effect, it won't change anything. But by not deleting them entireliy it makes it easier for Sheppy to come back to this macro and resume work on it, and it means you don't have to think about whether to delete the macro argument entirely.

tests/macros/utils.js
ExE-Boss: TypeScript is happy with it, also <code>return&nbsp;await &lt;promise&#x2011;result&gt;</code> [is&nbsp;generally not&nbsp;recommended](https://eslint.org/docs/rules/no-return-await).

Also, TypeScript is&nbsp;fine with&nbsp;it.

davidflanagan: I'm not sure I believe that there is a real performance issue with `return await`, but it seems reasonable enough to avoid it. I'm fine if you want to do an if/else and return a string from one branch and a promise from the other.  My objection is returning and || expression whose value is either a string or a promise. That seems like bad style and something that will cause confusion.

tests/macros/utils.js
ExE-Boss: I’ve added `readFixture` which doesn’t automatically add the file extension.

tests/macros/utils.js
ExE-Boss: I decided to add `unmockTemplate(…)` to do that.

tests/macros/utils.js
ExE-Boss: I like doing that too, but I somehow forgot about it this time.

tests/macros/InterfaceOverview.test.js
ExE-Boss: I’ve now added that using [`GlobalEventHandlers`](https://developer.mozilla.org/docs/Web/API/GlobalEventHandlers)

tests/macros/InterfaceOverview.test.js
wbamberg: As far as I can tell, external fixtures are only used in two tests at the moment:

* [compat.test.js](https://github.com/mdn/kumascript/blob/master/tests/macros/compat.test.js#L13-L20), which tests the Compat.ejs macro
* [dekiscript-page.test.js](https://github.com/mdn/kumascript/blob/master/tests/macros/dekiscript-page.test.js#L46-L53), added by @escattone in https://github.com/mdn/kumascript/pull/240, which I think is testing some Wiki-specific APIs that macros are able to use (specifically [`subpages`](https://github.com/mdn/kumascript/blob/master/src/environment.js#L871), [`subpagesExpand`](https://github.com/mdn/kumascript/blob/master/src/environment.js#L898), and so on.

These utility functions seem like they would be useful, yes. A common fixture for a lot of macros (especially sidebars) is going to be this list of page metadata that's returned by functions like `subPagesExpand` (because sidebars very often want to say "get me all the pages underneath this one" and then do something with that result):

https://github.com/mdn/kumascript/blob/master/macros/WebExtAPISidebar.ejs#L143
https://github.com/mdn/kumascript/blob/master/macros/APIRef.ejs#L44
https://github.com/mdn/kumascript/blob/master/macros/APIRef.ejs#L86
https://github.com/mdn/kumascript/blob/master/macros/ListSubpagesForSidebar.ejs#L29
https://github.com/mdn/kumascript/blob/master/macros/JSRef.ejs#L88

I don't know if there's any scope then for providing a generic utility to mock `subpagesExpand`, or something like that.

tests/macros/utils.js
```diff
@@ -154,11 +209,52 @@ function afterEachMacro(teardown) {
...
+ *        and an optional flag. If a flag is not provided, it defaults to `'r'`.
+ *
+ * @returns {string|Buffer}
+ */
+function readFixture(filePath, options) {
+    if (!Array.isArray(filePath)) {
+        filePath = [filePath];
+    }
+    let absolutePath = path.resolve(__dirname, 'fixtures', ...filePath);
+    if (options && typeof options !== 'string') {
```
escattone: I don't see the necessity for this chunk of code. Why not just pass `options` through to the `readFileSync` call? It seems that you're making a copy of the object, but I'm not sure why that's necessary?

ExE-Boss: It isn’t necessary, I just instinctively wrote this code.

ExE-Boss:
review?(@a2sheppy, @davidflanagan)

PR #1074, opened 2019-02-07
fix(macros): Correct {{DOMxRef(…)}} URL computation

### LINKED ISSUE #1054: domxref macros replaces only first " " to "_"
Example:
{{domxref("Fullscreen API")}} → https://developer.mozilla.org/en-US/docs/Web/API/Fullscreen_API (all OK)

{{domxref("Storage Access API")}} → https://developer.mozilla.org/en-US/docs/Web/API/Storage_Access%20API (broken link)

ExE-Boss:
Fixes #1054

review?(@davidflanagan)

### DIFF
--- a/macros/DOMxRef.ejs
+++ b/macros/DOMxRef.ejs
@@ -21,7 +21,10 @@ let str = $1 || $0;
 let rtlLocales = ['ar', 'he', 'fa'];
 let localStrings = string.deserialize(await template("L10n:Common"));
 
-api = api.replace(/ /g, '_').replace(/\(\)$|\(\)\./g, '').replace(/\.prototype\./g, '.').replace(/\./g, '/');
+api = api.replace(/ /g, '_')
+    .replace(/\(\)$|\(\)(?=\.)/g, '')
+    .replace(/\.prototype\./g, '.')
+    .replace(/\./g, '/');
 
 // RTL locales
 if (rtlLocales.indexOf(lang) != -1) {
--- a/macros/DOMxRef.ejs
+++ b/macros/DOMxRef.ejs
@@ -22,7 +22,7 @@ let rtlLocales = ['ar', 'he', 'fa'];
 let localStrings = string.deserialize(await template("L10n:Common"));
 
 api = api.replace(/ /g, '_')
-    .replace(/\(\)$|\(\)(?=\.)/g, '')
+    .replace(/\(\)/g, '')
     .replace(/\.prototype\./g, '.')
     .replace(/\./g, '/');
 
--- a/macros/DOMxRef.ejs
+++ b/macros/DOMxRef.ejs
@@ -5,13 +5,13 @@
  * a function.
  * @param
  *   The path of the page to link to, relative to the Web/API/ documentation path.
- * @param [optional] 
+ * @param [optional]
  *   The text to use for the link.  If omitted, the value of the first
  *   parameter will be used
- * @param [optional] 
- *   An anchor to link to on the page. Link text will display as $0.$2 or $1.$2 
- * @param [optional] 
- *   If set, do not put the domxref text in code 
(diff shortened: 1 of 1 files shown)


macros/DOMxRef.ejs
```diff
@@ -5,13 +5,13 @@
...
  *   The text to use for the link.  If omitted, the value of the first
  *   parameter will be used
- * @param [optional] 
- *   An anchor to link to on the page. Link text will display as $0.$2 or $1.$2 
- * @param [optional] 
- *   If set, do not put the domxref text in code 
+ * @param [optional]
+ *   An anchor to link to on the page. Link text will display as $0.$2 or $1.$2
+ * @param [optional]
+ *   If set, do not put the domxref text in code
```
davidflanagan: In general, I would prefer that you do not make minor formatting cleanups like this. What you're proposing is basically a one-line patch, and having these other changes is a distraction. Not a big deal, and no need to change it for this PR. You may have your editor configured to automatically remove trailing whitespace, but be aware that it makes reviews harder.

macros/DOMxRef.ejs
```diff
@@ -21,13 +21,14 @@ let str = $1 || $0;
 let rtlLocales = ['ar', 'he', 'fa'];
 let localStrings = string.deserialize(await template("L10n:Common"));
 
-api = api.replace(' ', '_').replace('()', '').replace('.prototype.', '.').replace(/\./g, '/');
+api = api.replace(/ /g, '_').replace(/\(\)$|\(\)\./g, '').replace(/\.prototype\./g, '.').replace(/\./g, '/');
```
davidflanagan: Please simplify the parens case to replace all '()' pairs, regardless of position in the string. Or convince me that it is a bug. Note however, that replacing '().' with '' is wrong: we'd want the dot to remain in that case.

I'd be fine just doing regular non-global string replacement for () and .prototype. as in the old code, since I can't imagine a case where they would appear twice. That would make your fix even more minimal and lower risk. But I'll accept the regex version, too.

macros/DOMxRef.ejs
ExE-Boss: This was indeed done by my editor (it trimmed the trailing whitespace)

macros/DOMxRef.ejs
```diff
@@ -21,13 +21,17 @@ let str = $1 || $0;
 let rtlLocales = ['ar', 'he', 'fa'];
 let localStrings = string.deserialize(await template("L10n:Common"));
 
-api = api.replace(' ', '_').replace('()', '').replace('.prototype.', '.').replace(/\./g, '/');
+api = api.replace(/ /g, '_')
+    .replace(/\(\)$|\(\)(?=\.)/g, '')
```
davidflanagan: Sorry to be a stickler about this, but since we don't have tests for this macro I don't want to risk replacing a very simple string replace with a complicated regexp like this.

I'll accept an unchanged `.replace('()', '')` here, or a simple regexp `.replace(/\(\)/g, '')`

If replacing all '()' that is not at the end of a line or before a period is a bug that need to be fixed here you'll have to convince me of that.

Your code is arguably the most correct way to do this, and it looks like it works. But it also seems more complicated than necessary, and we don't have tests, and I don't want to risk changing current behavior.

If someone has written a macro {{domxref('()foo')}}, this patch would change its output. That probably isn't the case, but why risk it?

ExE-Boss:
```suggestion
    .replace(/\(\)/g, '')
```

davidflanagan:
Thanks @ExE-Boss. I've merged. If you look at the git logs you'll see that I added a couple of sentences to the commit message. I like our git history to have verbose messages with a summary on the first line, but then a short paragraph or two describing the change in more detail.

PR #1078, opened 2019-02-15
test utility function for HTML linting with tests

escattone:
The main thrust of this PR is that it adds a `lintHTML()` test utility function -- which uses the [npm package](https://www.npmjs.com/package/vnu-jar) of the [Nu HTML Checker (v.Nu)](https://github.com/validator/validator) -- that can be used within the tests for many of the KumaScript macros (thanks to @wbamberg for that idea). I debated (and waffled) on whether I should make `lintHTML()` synchronous or asynchronous. I originally made it asynchronous, but decided in the end that it wasn't worth the extra mental/coding complexity given that its target environment was the test suite. I'd be happy to switch it back, or add it as an additional choice, if others disagree with that decision.

It also adds a new macro test (for the `AddonSidebar.ejs` macro) because I wanted to use `lintHTML()` at least once before submitting it for review. The tests for the `AddonSidebar.ejs` macro, however, go quite a bit further than just using `lintHTML()` (maybe too far, in the sense that it was a lot of work to understand and select an appropriate level of detail to test -- I'm hoping @wbamberg can guide me on that).

I also made two small tweaks to the `AddonSidebar.ejs` macro:
- I updated the Stack Overflow link to use `https` instead of `http`
- I stopped adding `closed` as an attribute to one of the `<details>` elements, since it's technically not a legal attribute for that element type and `lintHTML()` flags it as an error

This PR depends on https://github.com/mozilla/kuma/pull/5253 to add the JRE to the KumaScript Docker image, but since that has already been reviewed and merged, we're good to go there.

### DIFF
--- a/macros/AddonSidebar.ejs
+++ b/macros/AddonSidebar.ejs
@@ -9,7 +9,7 @@ function currentPageIsUnder(root) {
   if (slug && slug.indexOf(rootSlug) != -1) {
     return 'open';
   }
-  return 'closed';
+  return '';
 }
 
 var text = mdn.localStringMap({
@@ -439,7 +439,7 @@ var text = mdn.localStringMap({
         <ol>
           <li><a href="https://blog.mozilla.org/addons"><%=text['Blog']%></a></li>
           <li><a href="https://discourse.mozilla.org/c/add-ons"><%=text['Forums']%></a></li>
-          <li><a href="http://stackoverflow.com/questions/tagged/firefox-addon">Stack Overflow</a></li>
+          <li><a href="https://stackoverflow.com/questions/tagged/firefox-addon">Stack Overflow</a></li>
           <li><a href="<%=baseURL%>#Contact_us"><%=text['#Contact_us']%></a></li>
         </ol>
       </details>
--- a/tests/macros/addonsidebar.test.js
+++ b/tests/macros/addonsidebar.test.js
@@ -5,13 +5,15 @@ const url = require('url');
 
 const { JSDOM } = require('jsdom');
 
-const { beforeEachMacro,
-        describeMacro,
-        itMacro,
-        lintHTML } = require('./utils');
+const {
+    beforeEachMacro,
+    describeMacro,
+    itMacro,
+    lintHTML
+} = require('./utils');
 
 const SUMMARIES = {
-    'en-US': new Set([
+    'en-US': [
         'Getting started',
         'Concepts',
         'User interface',
@@ -24,12 +26,12 @@ const SUMMARIES = {
         'Publishing add-ons',
         'Distributing add-ons',
         'Channels'
-    ]),
-    'fr': new Set([
+    ],
+    fr: [
         'Démarrage',
         'Concepts',
         'Interface Utilisateur',
-        'Mode d\'emploi',
+        "Mode d'emploi",
         'Portage',
         'Déroulement avec Firefox',
         'Les API JavaScript',
@@ -38,8 +40,8 @@ const SUMMARIES = {
         'Publication de votre extension',
         'Distribuer votre module',
         'Canaux de discussions'
-    ]),
-    'ja': new Set([
+    ],
+    ja: [
         '始めましょう',
         '概念',
         'ユーザーインターフェイス',
@@ -52,7 +54,7 @@ const SUMMARIES = {
         'アドオンを公開する',
         'アドオンの配布',
         'チャンネル'
-    ]),
+    ]
 };
 
 const MANIFEST_SLUG = 'Mozilla/Add-ons/WebExtensions/manifest.json';
@@ -90,14 +92,14 @@ function getMockResultForFetchJSONResource(doc_url) {
                 subpages: [],
                 slug: `${MANIFEST_SLUG}/version`,
                 title: 'version'
-            },
+            }
         ],
         slug: MANIFEST_SLUG,
         title: 'manifest.json'
     };
 }
 
-function checkSidebarDom(html, locale, is_under_web_ext_api=false) {
+function checkSidebarResult(html, locale, isUnderWebExtAPI = false) {
     // Lint the HTML
     expect(lintHTML(html)).toBeFalsy();
     const dom = JSDOM.fragment(html);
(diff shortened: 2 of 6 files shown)


tests/macros/addonsidebar.test.js
```diff
@@ -0,0 +1,168 @@
+/**
+ * @prettier
```
davidflanagan: It looks like prettier has actually been run (with `npm run format`) on this file. But just want to check, because I often forget to do it, and we have not automated it at all.)

tests/macros/addonsidebar.test.js
```diff
@@ -0,0 +1,168 @@
...
+        section.querySelectorAll('ol > li.toggle').length
+    ).toEqual(SUMMARIES[locale].size);
+    // Check that all links reference the proper locale or use https
+    const num_total_links = section.querySelectorAll('a[href]').length;
+    const num_valid_links = section.querySelectorAll(
+        `a[href^="/${locale}/docs/Mozilla/Add-ons"], a[href^="https://"]`
+    ).length;
+    expect(num_valid_links).toEqual(num_total_links);
+    // Check that one of the list item's details are open by default (or not)
+    const details_open = section.querySelector(
```
davidflanagan: Might be better to use querySelectorAll() and then check the length of the returned array. That way you could also test that there was not more than one open?

tests/macros/addonsidebar.test.js
```diff
@@ -0,0 +1,168 @@
...
+                slug: `${MANIFEST_SLUG}/version`,
+                title: 'version'
+            },
+        ],
+        slug: MANIFEST_SLUG,
+        title: 'manifest.json'
+    };
+}
+
+function checkSidebarDom(html, locale, is_under_web_ext_api=false) {
```
davidflanagan: You're outing yourself as a Python programmer with all the underscores. :-)

I don't object to that style, and given that this is a test, I don't think it  is something you need to change.

tests/macros/addonsidebar.test.js
```diff
@@ -0,0 +1,168 @@
...
+        'ol > li.toggle > details[open]'
+    );
+    if (is_under_web_ext_api) {
+        expect(details_open).toBeTruthy();
+    } else {
+        expect(details_open).toBeFalsy();
+    }
+    // Check a sample of the DOM for localized content
+    for (const node of section.querySelectorAll('summary')) {
+        expect(SUMMARIES[locale].has(node.textContent)).toBe(true);
```
davidflanagan: If you made the summaries into ordinary arrays instead of sets, you could write `expect(SUMMARIES[locale]).toContain(node.textContent)`.

Just sayin', not a change that feels necessary.

tests/macros/addonsidebar.test.js
```diff
@@ -0,0 +1,168 @@
...
+
+describeMacro('AddonSidebar', function() {
+    beforeEachMacro(function(macro) {
+        // Mock calls to template('WebExtAPISidebar', [])
+        macro.ctx.template = jest.fn((macro, args) => {
+            // This template will be tested on its own, so nothing needed here.
+            return '';
+        });
+        // Mock calls to MDN.fetchJSONResource(doc_url)
+        macro.ctx.MDN.fetchJSONResource = jest.fn((doc_url) => {
```
davidflanagan: I think you don't need the arrow function here and can just pass getMockResultForFetchJSONResource as the argument to jest.fn()

tests/macros/utils.js
```diff
@@ -154,11 +158,44 @@ function afterEachMacro(teardown) {
...
+function lintHTML(html, fragment=true) {
+    if (fragment) {
+        html = `<!DOCTYPE html>
+                <html>
+                <head><title>test</title></head>
+                <body>${html}</body>
+                </html>`;
+    }
+    try {
+        execSync(`java -jar ${vnu} --errors-only --format text -`, {
```
davidflanagan: How long does it take to fire up the java process? If you start writing a lot of tests that use it, is it going to make the test suite slow? 

Is there a way to set this up to create the process on the first call and then keep it open and reuse it for subsequent calls? (I'm guessing that would work even though Jest runs tests in parallel... presumably there would be one Java process for each of Jest's child execution contexts, though I'm not really sure how that works.)

Not something to block on, but just wondering about the future of this.

I think you made the right decision keeping this synchronous... If it was async, you might end up with a bunch of tests all trying to start up java processes at the same time.

tests/macros/utils.test.js
```diff
@@ -0,0 +1,40 @@
...
+        title: 'with an illegal value for a link attribute',
+        html: '<a href="https://example.com" rel="xxx">an example</a>',
+        error: 'Bad value “xxx” for attribute “rel” on element “a”'
+    },
+];
+
+describe('test lintHTML function', function() {
+    for (let test of ERROR_TEST_CASES) {
+        it(test.title, function() {
+            expect(lintHTML(test.html)).toEqual(
```
davidflanagan: I think you can just use toContain() here

tests/macros/utils.test.js
```diff
@@ -0,0 +1,40 @@
...
+                expect.stringContaining(test.error)
+            );
+        });
+    }
+    it('with valid HTML input', async function() {
```
davidflanagan: You don't need the async, do you?

macros/AddonSidebar.ejs
```diff
@@ -9,7 +9,7 @@ function currentPageIsUnder(root) {
   if (slug && slug.indexOf(rootSlug) != -1) {
     return 'open';
   }
-  return 'closed';
+  return '';
```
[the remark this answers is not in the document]
escattone: Thanks for checking that! I had thought about that possibility after I quit last night, but failed to remember it this morning! 😃

tests/macros/addonsidebar.test.js
```diff
@@ -0,0 +1,168 @@
...
+    expect(
+        section.querySelectorAll('ol > li.toggle').length
+    ).toEqual(SUMMARIES[locale].size);
+    // Check that all links reference the proper locale or use https
+    const num_total_links = section.querySelectorAll('a[href]').length;
+    const num_valid_links = section.querySelectorAll(
+        `a[href^="/${locale}/docs/Mozilla/Add-ons"], a[href^="https://"]`
+    ).length;
+    expect(num_valid_links).toEqual(num_total_links);
+    // Check that one of the list item's details are open by default (or not)
```
wbamberg: As far as I can tell, this bit is testing specifically that if the slug is under 'WebExtensions/API' then the 'JavaScript APIs' summary is open, and that otherwise the summaries are all closed.

If this is right then I think it would be clearer to say that in this comment (and also perhaps in the test, to test that the open summary is in fact  'JavaScript APIs').

tests/macros/addonsidebar.test.js
```diff
@@ -0,0 +1,168 @@
...
+}
+
+describeMacro('AddonSidebar', function() {
+    beforeEachMacro(function(macro) {
+        // Mock calls to template('WebExtAPISidebar', [])
+        macro.ctx.template = jest.fn((macro, args) => {
+            // This template will be tested on its own, so nothing needed here.
+            return '';
+        });
+        // Mock calls to MDN.fetchJSONResource(doc_url)
```
wbamberg: It wasn't obvious to me that (from the point of view of AddonSidebar) this is mocking `tree` (https://github.com/mdn/kumascript/blob/master/macros/AddonSidebar.ejs#L390). I think it would be helpful to mention that here.

tests/macros/addonsidebar.test.js
escattone: That's for sure! 😃 I honestly didn't even think twice about this, and now I see how incongruent it is! I'll change this to camel case.

tests/macros/addonsidebar.test.js
escattone: Doh, missed that! Thanks!

tests/macros/utils.test.js
escattone: Good catch, cruft leftover from when I had `lintHTML()` as an asynchronous function.

tests/macros/utils.test.js
escattone: Oh? Cool, that's simpler so I'll give that a try.

tests/macros/addonsidebar.test.js
escattone: I have not run `prettier`, but I'll do that and see if anything changes, thanks!

tests/macros/addonsidebar.test.js
escattone: I like your idea better since if this fails, it'll produce a better error message.

tests/macros/utils.js
escattone: Thanks for recording your thoughts on this. I had thought of these issues as well, and come to the same conclusion. I suspect `lintHTML()` will be used heavily, so it's something we may have to consider soon. I've used [Python's built-in process pool](https://docs.python.org/2/library/multiprocessing.html#using-a-pool-of-workers) for things like this, and wondered if Node.js had something similar. I saw [this](https://www.npmjs.com/package/workerpool) npm package that looked promising.

tests/macros/utils.test.js
escattone: You're right on, thanks. Reading [this](https://jestjs.io/docs/en/expect#tocontainitem) again, I see that it's explicitly stated in the docs (`.toContain can also check whether a string is a substring of another string.`) which I missed when I first read it.

tests/macros/utils.js
sideshowbarker: > How long does it take to fire up the java process?

When vnu.jar is invoked from the command line, it takes about 2 seconds in total to start up the java process and then load the HTML5 schema and other init steps.

> Is there a way to set this up to create the process on the first call and then keep it open and reuse it for subsequent calls?

There’s no way to do that when invoking vnu.jar for command-line checking. There is a way to do it by instead starting the validator as a persistent web server, in which case there’s an HTTP/REST API that can be used to send documents to it one-by-one for checking and get the results back.

That avoids paying the java/init startup cost for each document request but has other tradeoffs — including that the exit code doesn’t automatically get set the way it does when invoking vnu.jar from the command-line (that is, if the validator finds any errors in a document when you check it from the command-line, when the process exits, it exits with a non-zero exit status — whereas if the document has no errors, it exits zero).

However, since, when invoking vnu.jar for command-line-checking, you can send documents to it in batches, you can also avoid paying the java/init startup cost by queuing/buffering up the documents and then feeding them to vnu.jar just once as a set.

But I guess that alternative may not be a good fit for the test harness here.

PR #1079, opened 2019-02-18
Update ToolsSidebar macro to remove obsolete extending devtools links

chrisdavidmills:
... and replace them with a link to the updated Browser Extensions tutorial

As per https://github.com/mdn/sprints/issues/1035

I have just made it link to the en-US version regardless of the locale, because there are not many translations of the page, but choosing the right one is a pain because the URL slugs are all localized too. So leaving it like this for now.


chrisdavidmills:
Cool, thanks @wbamberg

PR #1067, opened 2019-02-04
test(macros): Format GroupData test

ExE-Boss:
This formats the&nbsp;`GroupData` test using Prettier.

review?(@escattone, @wbamberg)

PR #1080, opened 2019-02-20
Basic tests for APIRef macro

escattone:
This PR adds some basic testing (HTML linting, basic `href` checks) for the `APIRef.ejs` macro. It also fixes an HTML error in `APIRef.ejs` found by the test. I've let all of the calls to `template` run without mocks, since they all effectively load the real data/HTML from files within the repo (they don't make any network calls).

~~I've added one additional, unrelated commit that updates `tests/macros/groupdata.test.js` with its `prettier` version, which had not yet been committed.~~

### DIFF
--- a/tests/macros/apiref.test.js
+++ b/tests/macros/apiref.test.js
@@ -1,0 +1,262 @@
+/**
+ * @prettier
+ */
+const { JSDOM } = require('jsdom');
+
+const {
+    beforeEachMacro,
+    describeMacro,
+    itMacro,
+    lintHTML
+} = require('./utils');
+
+const GAMEPAD_SUBPAGES_EXPAND = [
+    {
+        json_modified: '2019-01-16T23:02:17.747616',
+        subpages: [],
+        tags: [
+            'Property',
+            'Gamepad API',
+            'NeedsBetterSpecLink',
+            'Reference',
+            'NeedsMarkupWork',
+            'API',
+            'Référence(2)',
+            'Games'
+        ],
+        locale: 'en-US',
+        translations: [
+            {
+                uuid: '533ea9eb-06ae-4518-82bf-4f409ebae3d0',
+                title: 'Gamepad.axes',
+                url: '/ja/docs/Web/API/Gamepad/axes',
+                tags: [
+                    'API',
+                    'Gamepad API',
+                    'NeedsBetterSpecLink',
+                    'Reference',
+                    'NeedsMarkupWork',
+                    'Property',
+                    'Référence(2)',
+                    'Games'
+                ],
+                summary:
+                    '<a href="/ja/docs/Web/API/Gamepad" title="Gamepad API の Gamepad インターフェースはそれぞれのゲームパットやその他のコントローラーを定義し、ボタンのプッシュや軸位置やIDといった情報にアクセスできるようにします。"><code>Gamepad</code></a> インターフェースの <code><strong>Gamepad.axes</strong></code> プロパティは<span class="tlid-translation translation"><span title="">デバイス上に存在する軸を持つコントロールを表す配列を返します。</span></span> (例 : アナログスティック)。-',
+                localization_tags: [],
+                locale: 'ja',
+                last_edit: '2018-12-18T04:32:23.760574',
+                review_tags: []
+            }
+        ],
+        summary:
+            'The <code><strong>Gamepad.axes</strong></code> property of the <a href="/en-US/docs/Web/API/Gamepad" title="The Gamepad interface of the Gamepad API defines an individual gamepad or other controller, allowing access to information such as button presses, axis positions, and id."><code>Gamepad</code></a> interface returns an array representing the controls with axes present on the device (e.g. analog thumb sticks).-',
+        id: 87761,
+        review_tags: [],
+        slug: 'Web/API/Gamepad/axes',
+        uuid: 'c867795d-c9df-4919-b3ff-8f3809e0485a',
+        title: 'Gamepad.axes',
+        url: '/en-US/docs/Web/API/Gamepad/axes',
+        modified: '2019-01-16T13:03:20.906343',
+        label: 'Gamepad.axes',
+        localization_tags: [],
+        last_edit: '2018-07-20T12:20:59.066605',
+        sections: [
+            {
+                id: 'Quick_Links',
+                title: null
+            },
+            {
+                id: 'Syntax',
+                title: 'Syntax'
+            },
+            {
+                id: 'Example',
+                title: 'Example'
+            },
+            {
+                id: 'Value',
+                title: 'Value'
+            },
+            {
+                id: 'Specifications',
+                title: 'Specifications'
+            },
+            {
+                id: 'Browser_compatibility',
+                title: 'Browser compatibility'
+            },
+            {
+                id: 'sect1',
+                title: null
+            },
+            {
+                id: 'sect2',
+                title: null
+            },
+            {
+                id: 'sect3',
+                title: null
+            },
+            {
+                id: 'sect4',
+                title: null
+            },
+            {
+                id: 'sect5',
+                title: null
+            },
+            {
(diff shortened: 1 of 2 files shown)


ExE-Boss:
The prettier reformatting of&nbsp;`GroupData.test.js` should be&nbsp;left for&nbsp;https://github.com/mdn/kumascript/pull/1067.

escattone:
Thanks for pointing that out @ExE-Boss, I'll remove `tests/macros/groupdata.test.js` from this PR.

escattone:
Removed the `prettier` re-formatting of `tests/macros/groupdata.test.js` from this PR (that will be done by #1067).

wbamberg:
I think as far as it goes this is a good test, and it's obviously much better than the current no-test situation.

However it's quite minimal. I think that is OK with very static sidebar macros like [ToolsSidebar](https://github.com/mdn/kumascript/blob/a69bbb3822b6fc9d1f20e8124256a0af93847528/macros/ToolsSidebar.ejs), that just dump a blob of HTML into the page. But APIRef is a complex macro whose behaviour depends a lot on other sources and I don't think that this test would be enough for me to feel comfortable making changes to the code.

I don't want to suggest that the below is mandatory - I'm happy to merge this as it is - but just to give an outline of the sorts of things a comprehensive test might want to look at.

Also, and related, about mocking: the main value of mocking data sources, for me, is that it gives us a controlled way to test behaviour with a range of different inputs, including ones that are a bit pathological.

---------------------

## Overview of APIRef

This is a fairly high-level spec, as far as I can determine from the code.

### Inputs

* we get the `mainIF` from the slug of the page we are on. It's expected, I think, to be the name of a Web API interface, something like `Geolocation` or `Storage` or `Request`.

* we load InterfaceData.json, and use `mainIF` as a key to get the `inh` (inherited) and `impl` (implemented) arrays. We also expand these with some tricky logic in the `getInheritance` and `getImplementedBy` functions.

* we get the list of subpages underneath the `mainIF` Wiki page, and classify them according to their tags, into one of three buckets:
  * `ctors`
  * `properties`
  * `methods`


* the macro also accepts an optional string parameter: this is expected to be the name of an "API" (confusingly, in the Web API docs, APIs and interfaces are different things), like `Fetch API` or `Geolocation API`. This parameter is optional though and seems to be often omitted.

* we load GroupData.json, and use the parameter, if it was given, as a key to get:
  * interfaces, methods, and properties, which we stuff in an array called `related`
  * events, which we keep separately in an array called `events`

---------------------

So now we have the following things:

* arrays called `ctors`, `properties`, `methods` that we got from subpages of the "main interface page", and classified by the subpage tags

* arrays called `inh` and `impl` that we got from InterfaceData.json, based on the `mainIf`, which in turn was derived from the current page slug

* arrays called `related` and `events`, which we got from GroupData.json, based on the parameter we were given, if we were given one.

----------------------

### Outputs

We then build the sidebar like:

* list `ctors`, `properties`, `methods` in their own sections of the sidebar. We add little badges for items that are experimental, nonstandard, deprecated. If we are not in en-US, we also check whether subpages have been translated, and add a "call to action" if they haven't.

* list `inh`, `impl`, `events`, and `related` in their own sections of the sidebar.

-----------------------

## Possible tests

So we could devise tests for various things in here.

* say the current page is at /docs/Web/API/MainIF/, is `mainIF` correctly given as "MainIF"?

* say the current page is at /docs/Web/API/MainIF/something-else, is `mainIF` correctly given as "MainIF"?

* given a value of `mainIF` and a particular subtree under that page, are "ctors/"properties"/"methods" being handled correctly? e.g.

  * if there are no ctors, is the section empty? If there is one, is there one item? if there are two, are they both there?
  
  * if the subpages have tags for "experimental" &c, are the badges being shown.
  
  * there is some tricky string-fiddling in here (e.g. https://github.com/mdn/kumascript/blob/master/macros/APIRef.ejs#L140-L143) which always makes me nervous. Are there strings we might encounter that would break the HTML output?
  
  * is the CTA correctly added for non-en-US locales?

* given a parameter value to use for GroupData, check that "related" and "events" are calculated properly:

  * what if GroupData doesn't have an entry for the given parameter?
  
  * what if it has various different possible combinations of interfaces, methods, properties, events?
  
  * what if the names of these things contain weird characters? Will we break the HTML? (actually I know the answer to this: yes we will, but the GroupData test should ensure that doesn't happen. But perhaps APIRef should also be defensive.)

* given some different values for InterfaceData, do we calculate and show inherited and implemented correctly? Does the logic in `getInheritance` and `getImplementedBy` work properly, for different values?

In general I think we should pay attention to anywhere where there's string-fiddling going on, and/or where we're getting string data from somewhere and sticking it in HTML output.

escattone:
Thanks for your review @wbamberg! Your testing plan is really helpful!

> But APIRef is a complex macro whose behaviour depends a lot on other sources and I don't think that this test would be enough for me to feel comfortable making changes to the code.

I totally agree. My main goal was to get some basic testing in place given the time I had left to work on this during this sprint. This is such an involved macro that I think it probably deserves its own task (for another sprint) to get it into the kind of shape that we'd have confidence when making changes to the macro.

Is it worth merging as is? I could go either way. If we do merge this, we'd have to do so with the understanding that this doesn't yet solve the issue of creating a sufficient test for the `APIRef` macro, or perhaps it's best to just keep it open as a reminder that it needs to be fleshed-out more?

wbamberg:
> Is it worth merging as is? I could go either way.

Yes, me too. The worry is that people will see APIRef tests as "done". But I can also see it would be good to get a foundation in place for more extensive tests. I guess on balance I'd be in favour of merging it.

tests/macros/apiref.test.js
```diff
@@ -0,0 +1,262 @@
...
+                id: 'Legend',
+                title: 'Legend'
+            },
+            {
+                id: 'See_also',
+                title: 'See also'
+            }
+        ]
+    }
+];
```
davidflanagan: In one of ExeBoss's PRs that has not landed yet (I think) there is a utility function to read fixtures like this giant JSON blob from files. If this subpagesExpand data could be useful for other macro tests, consider putting it in a fixture file and reading it in with a utility function

tests/macros/apiref.test.js
```diff
@@ -0,0 +1,262 @@
...
+            return GAMEPAD_SUBPAGES_EXPAND;
+        });
+    });
+
+    for (const locale of ['en-US', 'fr', 'ja']) {
+        itMacro(`with locale ${locale}`, function(macro) {
+            macro.ctx.env.locale = locale;
+            macro.ctx.env.slug = 'Web/API/Gamepad';
+            return macro.call('Gamepad API').then(function(result) {
+                checkResult(result, locale);
```
davidflanagan: You've got two different tests combined into a single checkResult() here.

I think it would be clearer (though slower) to have two separate tests named tests:

```
describeMacro(..., ()=>{
   beforeEachMacro(macro => {
   });

    it('produces valid html', () => {
    });

    it('produces locale-specific links', () => {
    });
});
```

This would mean duplicating the `for locale of ` loop inside each it() function, and doubling the number of macro calls, but it would give you better reporting when failures occur.

davidflanagan:
I've added a couple of comments in the code, and would also suggest that William's extensive test plan be added to the code as a /* TODO */ comment so we don't lose those ideas.

ExE-Boss:
The PR adding those functions is&nbsp;https://github.com/mdn/kumascript/pull/1073, and I'd&nbsp:say it’s&nbsp;close to&nbsp;merging.

tests/macros/apiref.test.js
ExE-Boss: The PR adding those functions is #1073, and I'd say it’s close to merging.

ExE-Boss: I have extracted the `readJSONFixture(…)` helper function into #&lt;pr&gt;.

PR #1081, opened 2019-02-21
Add Screen Capture.

jan-ivar:


a2sheppy:
I forgot to add this when I added Screen Capture to GroupData. Whoops!

PR #1082, opened 2019-02-22
Add Web Share API - Level 2.

jpmedley:


jpmedley:
Is there anything I need to do to move this forward. I just submitted the same content a second time (see above) because this has been here long enough that I forgot about it.

jpmedley:
Is there anything I can do to move this along?

PR #1089, opened 2019-02-26
Add Deprecated tools link

irenesmith:
Adding a new page [Deprecated tools](https://developer.mozilla.org/en-US/docs/Tools/Deprecated_tools) to ToolsSidebar.ejs so that it will appear in the DevTools sidebar under "More Tools".

For reference, see: [ZenHub #820](https://app.zenhub.com/workspaces/mdn-sprint-board-5ab3d23fd0f4ea53b0022a61/issues/mdn/sprints/820)


irenesmith:
> 
> 
> Oops, that was supposed to be a request changes. :)

Chris helped me with the conflict but I see that this still isn't merged. Do you have to approve the changes?

PR #1091, opened 2019-03-02
Add Simpifiled Chinese

Crystal-RainSlide:
Had not check this with JsonLint, sorry if there is any error


Crystal-RainSlide:
Cite for the zh-CN translation of "Samsung Internet" => "三星浏览器":
https://www.samsung.com/cn/apps/samsung-internet/

All the other stuff should have nothing with citation.

PR #1092, opened 2019-03-04
Update status of W3C specs

dontcallmedom:


ExE-Boss:
You need to update the `spec2` test, which apparently depends on one of the W3C spec that was updated in&nbsp;this (I’ll probably make some mock data using #1093 once that gets merged).

PR #1077, opened 2019-02-14
docs: Add JSDoc to all built‑in library files

ExE-Boss:
I had originally opened #959 way back, but some of the JSDoc comments were converted to plain comments without type information in #1035, this restores all of them and adds new ones.

review?(@davidflanagan, @Elchi3, @wbamberg)

PR #1093, opened 2019-03-05
test(utils): Add `readFixture` and `mockTemplate` helper functions

ExE-Boss:
Extracted from #1073 and #1077, neither of which seem to be getting anywhere.

review?(@a2sheppy, @chrisdavidmills, @davidflanagan, @Elchi3, @wbamberg)


tests/macros/utils.js
```diff
@@ -1,86 +1,243 @@
...
+ * @param {T} x Actual value.
+ * @param {any} y Potential expected value.
+ */
 assert.equal = (x, y) => {
     expect(x).toEqual(y);
 };
 
 assert.eventually = {
+    /**
+     * Asserts non-strict equality (==) of actual and expected.
```
davidflanagan: This is not true. Jest's toEqual() matcher does deep equality comparison using Object.is()

tests/macros/utils.js
```diff
@@ -1,86 +1,243 @@
...
+ */
 assert.isTrue = value => {
     expect(value).toEqual(true);
 };
+
+/**
+ * Asserts that value is false.
+ *
+ * @template T Type of value.
+ * @param {T} value Actual value.
```
davidflanagan: boolean intead of T?

tests/macros/utils.js
```diff
@@ -1,86 +1,243 @@
...
 assert.notProperty = (value, prop) => {
     expect(value).not.toHaveProperty(prop);
 };
+
+/**
+ * Asserts that set1 and set2 have the same members. Order is not take into account.
+ *
+ * @template T Type of set values.
+ * @param {T[]} a1 Actual set of values.
+ * @param {T[]} a2 Potential expected set of values.
```
davidflanagan: I don't think this is correct. The types force the arguments to be arrays, but I'm pretty sure that the code allows any iterable values: including arrays, strings and sets. I wouldn't want typing here to limit the legal ways to use this function.

tests/macros/utils.js
```diff
@@ -1,86 +1,243 @@
...
+ */
 assert.sameMembers = (a1, a2) => {
     expect(new Set(a1)).toEqual(new Set(a2));
 };
 
+/**
+ * Asserts that haystack includes needle.
```
davidflanagan: You use "haystack" and "needle" in some places and "list" and "element" in others.

davidflanagan: nit: it would be a slightly more minimal diff if this function didn't move within the list of assertion functions.

tests/macros/utils.js
```diff
@@ -1,86 +1,243 @@
...
         url: 'https://developer.mozilla.org/'
     };
     let environment = new Environment(pageContext, templates, true);
+    let ctx = environment.prototypeEnvironment;
+
+    /** @type {Map<string, string>} */
+    const macroResults = new Map();
+    const realTemplate = ctx.template;
+    ctx.template = jest.fn(async (name, ...args) => {
+        let macroName = String(name)
```
davidflanagan: Why add the String() conversion? Are there cases where this function is being invoked with a non-string argument?  If so I think we'd want that to cause a test failure rather than be accepted.

tests/macros/utils.js
```diff
@@ -1,86 +1,243 @@
...
+            .replace(/:/g, '-')
+            .toLowerCase();
```
davidflanagan: Since this .replace().toLowerCase() sequence is used three times, maybe it would be good to define a utility function called normalizeMacroName(). (Still here inside of createMacroTestObject(), of course).

tests/macros/utils.js
```diff
@@ -183,20 +341,55 @@ function lintHTML(html, fragment = true) {
         });
         return null;
     } catch (error) {
-        const error_message = error.message
+        const error_message = /** @type {Error} */ (error).message
```
davidflanagan: I assume you added this comment because you're using a toolchain that complained about a posssible type error on this line. The problem is that unless you can convince the rest of the project to use the same tool, or to include a type checking step in the linting process, someone might just come along and remove the comment because they think that it hurts the readability of the code.

tests/macros/utils.js
```diff
@@ -183,20 +341,55 @@ function lintHTML(html, fragment = true) {
...
         return error_message;
     }
 }
 
+/**
+ * Reads a generic fixture file from the `fixtures` directory.
+ *
+ * @param {string|string[]} filePath
+ *        A path to a file relative to the `fixtures` directory.
+ * @param {string|{encoding?:string,flag?:string}} [options]
```
davidflanagan: I'd recommend just making this a string. I can't think of any case where readFileSync() would need anything other than the default "r" for the flag, can you?  If you're going to keep the option of using an object, maybe just refer tot he node docs in the comment. Because as it stands now you never actually say what the flag is for.

tests/macros/utils.js
```diff
@@ -183,20 +341,55 @@ function lintHTML(html, fragment = true) {
...
+ *
+ * @param {...string} filePath
+ *        A path to a file relative to the `fixtures` directory.
+ */
+function readJSONFixture(...filePath) {
+    let fileName = filePath.pop();
+    if (!path.extname(fileName)) {
+        fileName += '.json';
+    }
+    return JSON.parse(String(readFixture([...filePath, fileName])));
```
davidflanagan: nit: it would be more efficient to push the filename back onto the array than it is to allocate and fill a new array with ...

tests/macros/utils.test.js
```diff
@@ -26,13 +28,31 @@ const ERROR_TEST_CASES = [
...
         });
-    }
-    it('with valid HTML input', function() {
-        expect(lintHTML('<div>This is nice</div>')).toBeFalsy();
+    });
+
+    it('`readFixture` works', function() {
+        let result = String(readFixture('utils-test.txt'));
+
+        // prevent false positives from git.core.autocrlf on Windows
```
davidflanagan: Yuck. Instead how about just `result = result.trim()` on all platforms and then take out the newline in the toEqual() call below. That way you wouldn't need the platform dependency.

tests/macros/utils.test.js
```diff
@@ -26,13 +28,31 @@ const ERROR_TEST_CASES = [
...
+        it('with valid HTML input', function() {
+            expect(lintHTML('<div>This is nice</div>')).toBeFalsy();
         });
-    }
-    it('with valid HTML input', function() {
-        expect(lintHTML('<div>This is nice</div>')).toBeFalsy();
+    });
+
+    it('`readFixture` works', function() {
+        let result = String(readFixture('utils-test.txt'));
```
davidflanagan: I think it would be nice if the fixture files were in a utils-test/ subdirectory, to keep the fixtures directory a bit cleaner. And that would exercise the case where there is more than one argument to readFixture.  Maybe utils-test/fixture.txt and utils-test/fixture.json or something like that?

tests/macros/fixtures/utils-test.json
```diff
@@ -0,0 +1,3 @@
+{
```
davidflanagan: See below where I propose slightly different names for these two fixture files.

davidflanagan (changes requested):
@ExE-Boss: thanks for pushing this. I continue to think that these are really useful test utilities. I think I had a couple of nits, but overall I think they're well written and it is great that you've included tests.

My request for changes has to do with all the type comments you've added. There are a couple of errors in the comments that will need to be fixed at a minimum. But there's also the larger issue of whether we want to be adding this kind of type information to this code. I don't have much of an opinion myself, but I'm not sure it makes sense to go to the effort of adding all this information unless there is consensus among the active contributors (and I don't include myself in that list) that the type information adds enough value to be worth it.

That is maybe a question for @wbamberg and @escattone?

tests/macros/utils.js
escattone: I personally don't like this kind of thing at all. I just don't find it readable or helpful.

tests/macros/utils.doc.js
```diff
@@ -0,0 +1,35 @@
+// TODO: Consider using a TypeScript definition file for this.
```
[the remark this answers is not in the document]
escattone: I would vote for removing this file, and adding these comments into `utils.js`.

wbamberg:
> My request for changes has to do with all the type comments you've added. 
> That is maybe a question for @wbamberg and @escattone?

My view: I don't obviously see a need for these, no. But more strongly, I'd very much prefer to see a single PR for a single purpose. This PR is to add the helper functions, so I'd prefer if it did just that. Then we could make the type comments a separate discussion. I think this is good as a general policy: it's easier for reviewers, and it's easier to have clear conversations and decisions about what we want to have in the repo, and to prioritise this work too.

escattone:
I totally agree with @wbamberg and @davidflanagan, but wanted to add one more thing. I'm not a fan of the type comments, and would personally prefer that they were not added. In my opinion, they're a burden and add little, if any, insight to that offered by the prose comments (which I do like).

tests/macros/utils.js
ExE-Boss: I mostly copied the type definition from chai, since the `assert` utility is meant as a compatibility shim between chai and Jest.

tests/macros/utils.js
ExE-Boss: I copied the description for this from the node docs.

ExE-Boss:
While working on&nbsp;this, I&nbsp;discovered an&nbsp;unrelated issue which I’m&nbsp;fixing in&nbsp;https://github.com/mdn/kumascript/pull/1057#pullrequestreview-211431973.

tests/macros/utils.js
```diff
@@ -4,83 +4,269 @@
...
+ */
 assert.notProperty = (value, prop) => {
     expect(value).not.toHaveProperty(prop);
 };
+
+/**
+ * Asserts that set1 and set2 have the same members. Order is not take into account.
+ *
+ * @template T Type of set values.
+ * @param {Iterable<T>} a1 Actual set of values.
```
davidflanagan: I suspect that Iterable<any> might be better here.

tests/macros/utils.js
```diff
@@ -4,83 +4,269 @@
...
+ * @remarks The assertion does match subclassed objects.
+ */
 assert.isObject = value => {
     expect(value).toBeInstanceOf(Object);
 };
+
+/**
+ * Asserts that value is a function.
+ *
+ * @param {Function} value Actual value.
```
davidflanagan: So this raises interesting questions....  The point of a runtime assertion that something is a function presumably means that we don't know statically whether it is a function. So declaring the argument to statically be a function seems like it will cause trouble...

It doesn't look like you've answered my questions about who or what these comments are for. If these are just really detailed doc comments for programmers reading the code, then this is probably fine. But if some developers will be actually using type-checking tools that read these comments, then I worry this will cause problems for tests that use assert.isFunction

tests/macros/utils.js
```diff
@@ -4,83 +4,269 @@
...
+     */
     async equal(x, y) {
         expect(await x).toEqual(y);
     }
 };
 
+/**
+ * Asserts that list includes the supplied element.
+ *
+ * @template T Type of values in list.
```
davidflanagan: The elements of the list don't all have to be the same type, so maybe better here to use any instead of T?

davidflanagan (changes requested):
I like this version much better than the last. I'm requesting changes to the default encoding of readFixture(), and I think that there might still be a few problems with types for the assertion functions. 

Also: I don't think you've actually explained the point of the comments. I've got enough experience with Flow to find these comments readable and kind of helpful, though I suspect that not all developers would agree with that. I assume that you're using these yourself with vscode as described in https://medium.com/@trukrs/type-safe-javascript-with-jsdoc-7a2a63209b76 Is that right?

I'm really not sure how I feel about that. I really like Flow and I understand the benefits of typed JavaScript. But I don't know what is going to happen when some contributors to a project are using type checking and some are not.  Basically, I worry that changes to this code will break the types, and then you'll have vscode reporting errors to you that no one else sees or cares about. And if you then submit PRs to fix them, we'll be taking time to review those PRs. Basically: I like typed JS, but I fear that in this case it will slow us down, and that for this relatively small codebase it is not really worth adding them.

On the other hand, if the type information is just there for human readers and not for automated tools, then it seems fine to add this extra detail in the comments.

Finally, I'd really like to see a commit message that is at least a paragraph long explaining the changes in the PR. This is so that we get useful information from `git log` once we merge.

tests/macros/utils.test.js
```diff
@@ -26,13 +26,25 @@ const ERROR_TEST_CASES = [
...
+        it('with valid HTML input', function() {
+            expect(lintHTML('<div>This is nice</div>')).toBeFalsy();
         });
-    }
-    it('with valid HTML input', function() {
-        expect(lintHTML('<div>This is nice</div>')).toBeFalsy();
+    });
+
+    it('`readFixture` works', function() {
+        let result = String(readFixture('utils-test/fixture.txt')).trim();
```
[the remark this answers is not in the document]
ExE-Boss: Yes.

I&nbsp;use&nbsp;Windows, so&nbsp;I&nbsp;know that my&nbsp;code works on&nbsp;Windows.

tests/macros/utils.js
ExE-Boss: In that case, I’ll convert it all to `any`.

tests/macros/utils.js
```diff
@@ -4,83 +4,269 @@
...
 assert.equal = (x, y) => {
     expect(x).toEqual(y);
 };
 
 assert.eventually = {
+    /**
+     * Asserts deep equality of actual and expected.
+     *
+     * @template T Type of the resolved value.
+     * @param {T|Promise<T>} x A promise which resolves to the actual value.
```
ExE-Boss: The type parameter has to be specified here, otherwise the&nbsp;type would collapse into&nbsp;`any`, instead&nbsp;of&nbsp;`T|Promise<T>`.

tests/macros/utils.js
ExE-Boss: I’ve set the list to&nbsp;`string|any[]`, but the&nbsp;`element` has&nbsp;to&nbsp;be&nbsp;`string|T`, otherwise the result would collapse into&nbsp;`any`.

tests/macros/utils.js
```diff
@@ -191,12 +385,54 @@ function lintHTML(html, fragment = true) {
...
+/**
+ * Reads a JSON fixture file from the `fixtures` directory.
+ *
+ * @template T The schema of the JSON payload.
+ * @param {...string} filePath
+ *        A path to a file relative to the `fixtures` directory.
+ *
+ *        The `.json` file extension can be omitted.
+ *
+ * @returns {T} The parsed JSON payload.
```
ExE-Boss: This&nbsp;is&nbsp;added so&nbsp;that the&nbsp;type cast gets performed automatically.

tests/macros/utils.js
```diff
@@ -191,12 +385,54 @@ function lintHTML(html, fragment = true) {
...
 
+/**
+ * Reads a generic fixture file from the `fixtures` directory.
+ *
+ * @param {string|string[]} filePath
+ *        A path to a file relative to the `fixtures` directory.
+ * @param {string|null} [encoding]
+ *        The file encoding. Set to `null` to load the file as a binary `Buffer`.
+ *
+ * @returns {string}
```
ExE-Boss: Unfortunately, JSDoc doesn’t provide any way to define function overloads, so&nbsp;the&nbsp;return type will always show up&nbsp;as&nbsp;a&nbsp;`string`, unless we create `utils.d.ts` or&nbsp;migrate this file to&nbsp;TypeScript.

PR #1097, opened 2019-03-12
Adding dictionaries for WebAuthn / Web Authentication API

SphinxKnight:
Adding both:
*  [`PublicKeyCredentialRequestOptions`](https://developer.mozilla.org/en-US/docs/Web/API/PublicKeyCredentialRequestOptions) 
* [`PublicKeyCredentialCreationOptions`](https://developer.mozilla.org/en-US/docs/Web/API/PublicKeyCredentialCreationOptions)
as dictionaries relevant to the Web Authentication API sidebar.
This is part of the ongoing work for https://github.com/mdn/sprints/issues/976

### DIFF
--- a/macros/GroupData.json
+++ b/macros/GroupData.json
@@ -1895,7 +1895,9 @@
                             "AuthenticatorAssertionResponse" ],
             "methods":    [],
             "properties": [],
-            "events":     []
+            "events":     [],
+            "dictionaries": ["PublicKeyCredentialRequestOptions",
+                             "PublicKeyCredentialCreationOptions"]
         },
         "Web Components": {
             "guides":       [


SphinxKnight:
@chrisdavidmills (+ @wbamberg ?) it has been more than 2 months since I have filed this. Is there anything I can do to help with merging this 3 lines patch?

chrisdavidmills:
Hi @SphinxKnight ! As luck would have it, I managed to get my local Kuma install working again at the end of last week. I've pulled your branch into my kumascript submoduile and tested it.

So, I am a little confused. The checks have passed, and your changes don't seem to create any errors, which is a good sign.

However, as you can see in the attached screenshot showing the output of `{{APIRef("Web Authentication API")}}`, the dictionary entries don't appear. One of the interfaces is missing too. They don't when I test `{{DefaultAPISidebar("Web Authentication API")}}` either, although the missing interface appears when I run that one.

@wbamberg do you know more about this? I thought we had code to display the dictionaries now? Am I wrong?

wbamberg:
>  I thought we had code to display the dictionaries now?

No, there is https://github.com/mdn/kumascript/pull/738 but it stalled on not having tests.

wbamberg:
But anyway we should accept this PR IMO: GroupData and all its major dependencies now have tests, and the changes look good to me.

chrisdavidmills:
OK, thanks for clarifying Will. I am OK to merge this, in that case.

SphinxKnight:
Thanks a lot @wbamberg & @chrisdavidmills!

PR #1099, opened 2019-03-19
refactor: Use extensionless require

ExE-Boss:
This converts the project to use extensionless require, like what is being done in tests.

review?(@davidflanagan)

### DIFF
--- a/run.js
+++ b/run.js
@@ -15,8 +15,8 @@ if (process.env.NEW_RELIC_LICENSE_KEY && process.env.NEW_RELIC_APP_NAME) {
 }
 
 // Start up a server instance.
-const config = require('./src/config.js');
-const Server = require('./src/server.js');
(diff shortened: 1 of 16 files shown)


davidflanagan:
I prefer using extensions for local modules, with both require and import syntax because when I'm reading code, if I want to go look at another module, it makes it very clear which file to open. If we don't use extensions then you can get into unexpected behavior if you inadvertently create a .js and a .jsx or .json file with the same base name.

If you'd like to argue for removing the extensions, I'm willing to be convinced. But if the only reason is for consistency with the tests, I'd prefer to edit the tests to match the src/ directory rather than change these files.

ExE-Boss:
The main reason is to ease migration to TypeScript (which uses the `.ts` extension).

I’ve already got a working branch in my repository which passes all the current tests and seems to work fine.

PR #1100, opened 2019-03-19
refactor(env): Extract Prototypes into separate files

ExE-Boss:
This extracts all the `*Prototype` objects into separate files in `src/api/*`, which will make testing easier.

## To do:
- [x] `util`
- [x] `mdnPrototype`
- [x] `stringPrototype` (I’ll also apply my fixes from #1013)
- [x] `wikiPrototype`
- [x] `webPrototype`
- [x] `pagePrototype`

### Simple re‑exports:
I’m considering leaving these ones in `environment.js`, as they just re‑export functions:
- [x] ~~`globalsPrototype`~~ (reverted&nbsp;in&nbsp;https://github.com/mdn/kumascript/pull/1100/commits/279561a25310c588a41d2db96987dcf985d1f5a2)
- [x] `kumaPrototype`
- [x] `uriPrototype`

---

I’m also adding JSDoc comments from #1077.

review?(@davidflanagan, @escattone)


ExE-Boss:
The issue with `globalsAPI` is that it exposes the unmodified `require(…)` function.

By moving it to the `api` directory, this would change the behaviour of relative `require(…)`.

I’ve modified the code so that relative `require(
)`

src/api/uri.js
```diff
@@ -0,0 +1,19 @@
...
+/**
+ * Encode text as a URI component.
+ *
+ * FIXME: This actually calls `encodeURI` instead of `encodeURIComponent`.
+ *
+ * @param {string} str
+ * @return {string}
+ */
+function encode(str) {
+    return encodeURI(str);
```
ExE-Boss: This should probably be changed to call `encodeURIComponent(…)` instead.

src/api/wiki.js
```diff
@@ -0,0 +1,333 @@
...
+        var tz = [],
+            x = 0,
+            y = -1,
+            n = 0,
+            i,
+            j;
+
+        while ((i = (j = t.charAt(x++)).charCodeAt(0))) {
+            var m = i == 46 || (i >= 48 && i <= 57);
+            if (m !== n) {
```
ExE-Boss: error TS2367: This condition will always return 'true' since the types 'boolean' and 'number' have no overlap.

src/api/wiki.js
```diff
@@ -0,0 +1,333 @@
...
+                tz[++y] = '';
+                n = m;
```
ExE-Boss: error TS2322: Type 'boolean' is not assignable to type 'number'.

src/api/wiki.js
```diff
@@ -0,0 +1,333 @@
...
+
+    function alphanumForward(a, b) {
+        var aa = chunkify(a.title);
+        var bb = chunkify(b.title);
+
+        for (let x = 0; aa[x] && bb[x]; x++) {
+            if (aa[x] !== bb[x]) {
+                var c = Number(aa[x]),
+                    d = Number(bb[x]);
+                if (c == aa[x] && d == bb[x]) {
```
ExE-Boss: error TS2367: This condition will always return 'false' since the types 'number' and 'string' have no overlap.

src/api/wiki.js
```diff
@@ -0,0 +1,333 @@
...
+
+    function alphanumBackward(a, b) {
+        var bb = chunkify(a.title);
+        var aa = chunkify(b.title);
+
+        for (let x = 0; aa[x] && bb[x]; x++) {
+            if (aa[x] !== bb[x]) {
+                var c = Number(aa[x]),
+                    d = Number(bb[x]);
+                if (c == aa[x] && d == bb[x]) {
```
ExE-Boss: error TS2367: This condition will always return 'false' since the types 'number' and 'string' have no overlap.

src/api/globals.js
```diff
@@ -0,0 +1,15 @@
...
+
+module.exports = {
+    /**
+     * #### require(name)
+     *
+     * Load an npm package (the real "require" has its own cache).
+     */
+    // TODO: Consider supporting macros in sub-directories
+    require: util.createRequire(config.macrosDirectory),
+};
```
davidflanagan: Oh, this is interesting that you have to do it that way. In that case, I'd prefer to just define globalsAPI directly in environment.js so that we don't need this special utility function and to reduce the risk of the change.  Sorry to ask you to change it and then change it back.

The risk of subtle behavior differences between the "real" require and the function returned by createRequire() is what I worry about, since we don't have a robust test suite.

src/api/uri.js
davidflanagan: I don't think it would be safe to make that change. It probably should have been written that way to begin with, but any change now risks changing what appears on developer.mozilla.org and I don't want to risk that since we don't have a way to test that API changes here don't affect the output on any of the thousands of MDN document, I'm very very reluctant to change anything.

src/api/wiki.js
davidflanagan: I wonder if this code was a port of a C algorithm, and `n=0` above was intended to be `n=false`? Maybe the !== should have been a !=

But again, unless we can point to a specific content bug that is being caused by this, I don't want to go changing this legacy code for fear of causing unexpected changes to the output.

src/api/wiki.js
davidflanagan: That's just not true. `(1 == "1")` evaluates to true in JavaScript

davidflanagan: It looks like this code is so old and untyped that it is not possible to apply type analysis to it without a complete rewrite. And rewrites are off the table.

ExE-Boss:
I haven’t actually changed any code, the only thing I’ve done is add `TODO` and `FIXME` comments.

davidflanagan:
I just saw that you pushed another change... Let me know when you're done and I'll take a final look and merge. I really would prefer that you put the '.js' extensions back on the require() calls, though.

ExE-Boss:
I’d say I’m done now.

davidflanagan:
I just noticed what you said above about keeping createRequire() in utils with a TODO comment. Please don't do that. I'd like to put KumaScript into maintenance mode and I don't think it will be worth while to make that kind of change in the future.

And again, please add '.js' extensions to your require() calls, or try to convince me why we shouldn't be using them.  My argument for using extensions is that without them the code is less clear and it is not clear what would happen if you were to drop a wiki.json or wiki.ts file into the api/ directory. I'd prefer to avoid code that depends on the precise behavior of the module loader and the contents of directories.

ExE-Boss:
Well, I guess I can always just change them to `.ts` as part of the conversion to TypeScript.

src/api/util.js
```diff
@@ -0,0 +1,210 @@
+/**
+ * Utility functions are collected here. These are functions that are used
+ * by the exported functions below. Some of them are themselves exported.
+ *
+ * @prettier
```
davidflanagan: This file is indented with tabs, which means that prettier has not been run on it. Either you forgot to run it before committing, or this @prettier tag here needs to be in a comment of its own up above, as it is in the other files.  Please reformat.

src/api/util.js
```diff
@@ -0,0 +1,210 @@
...
+/**
+ * Creates a `require` function wrapper that resolves
+ * relative requires against the macro directory.
+ *
+ * TODO: Consider supporting macros in sub-directories.
+ *
+ * @param {string} macroPath The path of the macro
+ * @return {NodeRequireFunction}
+ */
+function createRequire(macroPath) {
```
davidflanagan: Please remove this function since it isn't being used at this point.

davidflanagan:
I was hoping to merge this today, but in doing a final check, I found some more changes I'd like you to make. A couple are just nits, but one is serious.

1) `npm run format` makes a couple of formatting changes to environment.js. Please fix that. (It also changes a test file that you have not touched, but feel free to fix that one as well.)

2) In Makefile and package.json there are targets for test-coverage, but they only scan src/*, and now that src has a subdirectory they need to be updated to src/**, I think.

3) In src/api/strings.js, you change the casing of the functions. For example, you changed ToUpperFirst() to toUpperFirst(). I agree that the original names are horrible and ugly, but since we have no good way to test changes here, we have to be completely paranoid about compatibility.  The prepareProto() function messes with case, but not equivalently. ToUpperFirst was available as ToUpperFirst() and as toupperfirst(), but it was never available as toUpperFirst(), so by adding that you've introduced a new function into the API. That is not likely to break things, but it could, and I don't want to risk it. So please revert these functions to their original uppercase form.

4) When I was imagining moving these chunks of code into src/api, what I envisioned was files that looked like this:

```
// a couple of requires, as needed

module.exports = /* an object literal copied verbatim from environment.js */;
```

Environment.js already has the APIs defined as object literals with methods inside them, and I was imagining that you could use those object literals exactly as they are with the nice method syntax and without having to add the `function` keywords. Doing that would make it clearer that nothing had changed and that the refactor was completely safe.  As it is, the formatting and indentation changes enough that it is harder to see what has changed.

I would prefer it if these api files were in that `module.exports = {big object with methods};` form. But if you feel like that would be too much work to do that conversion at this point, I'm willing to accept them as-is.

ExE-Boss:
Turns out, the issue was trailing whitespace, which somehow didn’t get removed by my IDE.

And also a trailing comma, which were used in many places before #1035.

---

Also, I did a search, and nothing depends on the existence of `lowerCamelCase` versions of the `stringAPI` namespace methods, except for when I write my code, and then notice that it doesn’t work, because `prepareProto(…)` assumes that the default method name is in the `lowerCamelCase` variant, and so it doesn’t generate one.

davidflanagan:
The thing about the string API is that it is a horrible API that macro writers would be better off not using. Giving the functions better names will only make the API more attractive to use, which is the opposite of what we want.

The prepareProto() function doesn't make any assumptions about the method name beginning with a lower case letter: it simply matches the legacy behavior of the old KumaScript. Since we're not actively working on KumaScript anymore, I really don't want any risk of subtle changes that could cause future maintenance issues.

Changing the case of those function names converts this PR from a pure refactor with no behavior changes into something else. We'll have to get input from other team members if you want to proceed with the lower-case variants.

src/api/mdn.js
```diff
@@ -132,7 +132,7 @@ function replacePlaceholders(string, replacements) {
         /** @type {string | number} */
         var index = placeholder.substring(1, placeholder.length - 1);
         if (!Number.isNaN(Number(index))) {
-            index = Number(index) - 1;
+            index--;
```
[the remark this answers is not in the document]
ExE-Boss: This was how it was written before.

src/api/mdn.js
```diff
@@ -147,8 +147,8 @@ function replacePlaceholders(string, replacements) {
  * @param {string|{url:string}} fileObjOrUrl
  */
 async function getFileContent(fileObjOrUrl) {
-    var file_url =
-        typeof fileObjOrUrl === 'string' ? fileObjOrUrl : fileObjOrUrl.url;
+    /** @type {string} */
+    var file_url = fileObjOrUrl.url || fileObjOrUrl;
```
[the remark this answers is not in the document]
ExE-Boss: Ditto:
https://github.com/mdn/kumascript/blob/f0e471ef786dbe5c019c3da037b5c45f002d95cf/src/environment.js#L293-L297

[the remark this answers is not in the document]
ExE-Boss: Also, this version is type unsafe.

src/api/mdn.js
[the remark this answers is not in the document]
ExE-Boss: The thing is that the `index--` to `index = Number(index) - 1` snuck into this as part of my type safe improvements from my TypeScript branch.

ExE-Boss:
Ok, I’ll make a new PR where I don't make any changes, and then rebase this on top of it.

ExE-Boss:
I’ve now opened #1103 to do that.

PR #1103, opened 2019-03-28
refactor(env): Extract Prototypes into separate files

ExE-Boss:
A single atomic version of #1100 without JSDoc or type information from #1077.

This extracts all the `*Prototype` objects into separate files in `src/api/*`, which will make testing easier.

review?(@davidflanagan)


package.json
```diff
@@ -14,10 +14,10 @@
     "scripts": {
         "start": "node run.js",
         "test": "jest",
-        "test-coverage": "jest --coverage --collectCoverageFrom='src/*'",
+        "test-coverage": "jest --coverage --collectCoverageFrom='src/**'",
         "lint": "eslint *.js src tests",
         "lint-json": "find . -name '*.json' -not -path './node_modules/*' -print0 | xargs -t -n1 -0 jsonlint -q",
-        "format": "./node_modules/.bin/prettier --write --require-pragma *.js src/*.js tests/*.js tests/macros/*.js"
+        "format": "./node_modules/.bin/prettier --write --require-pragma *.js src/*.js src/**/*.js tests/*.js tests/**/*.js"
```
davidflanagan: The `**` thing isn't working quite right: this now omits src/*.js and tests/*.js. I've added my own mini commit to change htat.

davidflanagan:
Thank, you ExE-Boss!

I "diffed the diffs" to verify that there are no code changes here: this patch just moves chunks of environment.js into api/*.js and replaces them with requires() calls.

I tweaked the `npm run format` command in package.json with a tiny commit of my own and verified that all of the JS files have been correctly formatted with prettier.

I ran the tests locally, and Docker ran them as well: they're all passing. 

And I tested on my dev server to verify that Cmd-Shift-R still successfully re-renders the macros on a page.

davidflanagan:
Note that I edited the commit message when I squashed and merged to provide more detail about the purpose and content of the patch.

PR #1106, opened 2019-03-29
Bug 1556549: Add missing `.icon‑only‑inline` classes

ExE-Boss:
Part of [bug 1556549].

This was missed in #981 by @schalkneethling.

I’ve also renamed the `deprecated_inline.test.js` file, so that `git blame` information won’t be destroyed by merging #984.

---

[bug 1556549]: https://bugzilla.mozilla.org/show_bug.cgi?id=1556549

review?(@escattone, @wbamberg)
review?(@davidflanagan): Since you don’t seem to be merging #983 or #984, which would achieve this.

### DIFF
--- a/macros/DeprecatedBadge.ejs
+++ b/macros/DeprecatedBadge.ejs
@@ -18,7 +18,7 @@ var str = mdn.localString({
 });
 
 if ($0) {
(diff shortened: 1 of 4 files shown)


schalkneethling:
@escattone Can you merge? Thanks!

PR #1116, opened 2019-04-16
Update IndexedDB 2 macro

UziTech:
`"Indexed Database API"` points to IndexedDB 2.0 spec and
`"Indexed Database API 2.0"` point to IndexedDB 3.0 spec

I changed `"Indexed Database API 2.0"` to `"Indexed Database API Draft"` since it points to the editor's draft.

https://www.youtube.com/watch?v=y1TEPQz02iU&feature=youtu.be&t=541

### DIFF
--- a/macros/SpecData.json
+++ b/macros/SpecData.json
@@ -875,17 +875,12 @@
     "status": "PR"
   },
   "IndexedDB": {
-    "name": "Indexed Database API",
+    "name": "Indexed Database API 2.0",
     "url": "https://www.w3.org/TR/IndexedDB/",
     "status": "REC"
   },
-  "IndexedDB Draft": {
-    "name": "Indexed Database API Draft",
-    "url": "https://w3c.github.io/IndexedDB/",
-    "status": "Draft"
-  },
   "IndexedDB 2": {
-    "name": "Indexed Database API 2.0",
+    "name": "Indexed Database API Draft",
     "url": "https://w3c.github.io/IndexedDB/",
     "status": "REC"
   },
--- a/macros/SpecData.json
+++ b/macros/SpecData.json
@@ -879,10 +879,10 @@
     "url": "https://www.w3.org/TR/IndexedDB/",
     "status": "REC"
   },
-  "IndexedDB 2": {
-    "name": "Indexed Database API 2.0",
+  "IndexedDB Draft": {
+    "name": "Indexed Database API Draft",
     "url": "https://w3c.github.io/IndexedDB/",
-    "status": "REC"
+    "status": "Draft"
   },
   "InputDeviceCapabilities": {
     "name": "InputDeviceCapabilities",
--- a/macros/SpecData.json
+++ b/macros/SpecData.json
@@ -884,6 +884,11 @@
     "url": "https://w3c.github.io/IndexedDB/",
     "status": "Draft"
   },
+  "IndexedDB 2": {
+    "name": "Indexed Database API 2.0",
+    "url": "https://w3c.github.io/IndexedDB/",
+    "status": "REC"
+  },
   "InputDeviceCapabilities": {
     "name": "InputDeviceCapabilities",
     "url": "https://wicg.github.io/InputDeviceCapabilities/",


UziTech:
can this get reviewed?

wbamberg:
Sorry to be slow.

This data is used in places like https://developer.mozilla.org/en-US/docs/Web/API/indexeddb_api#Specifications, where a link is rendered in the page by calling some macro code with the name of the property in SpecData.json, like this:

```
{{SpecName("IndexedDB 2")}}
{{SpecName('IndexedDB')}}
```

(see https://developer.mozilla.org/en-US/docs/Web/API/IndexedDB_API$edit for these calls)

So we can't just rename "IndexedDB 2" or those pages will break.

But as you say, the data just seems wrong at the moment.

The implication (from https://developer.mozilla.org/en-US/docs/Web/API/indexeddb_api#Specifications) seems to be that 'IndexedDB' should refer to the initial version of the IndexedDB spec, and 'IndexedDB 2' should refer to 2.0.

So one option is: fix the links in  'IndexedDB' and 'IndexedDB 2' to point to the correct versions, add a new 'IndexedDB 3', and correspondingly add an extra row in all the IndexedDB spec tables to list it.

But honestly, it feels foolish to me to list every version of every spec, and pile them up in spec tables like this. It just feels like busy work - someone who cares enough to look at the spec link can find other versions for themselves. So I'd suggest instead to remove all except 'IndexedDB', make that point at the latest version, and update all pages to refer only to that spec.

However I'm not that clear on the our policy regarding spec tables at the moment. I know we had a conversation about them recently (not publicly, unfortunately) but I'm not sure what the conclusion is.

Either way it would be good to have clear guidance here. @chrisdavidmills , do you know what the status of this work is?

UziTech:
> it feels foolish to me to list every version of every spec

I agree, which is why I changed `IndexedDB 2` to `IndexedDB Draft` so we can update where `IndexDB` and `IndexDB Draft` point to without adding new versions in the future.

`IndexDB` will point to the latest released version and `IndexDB Draft` will point to the latest draft version.

This will break https://developer.mozilla.org/en-US/docs/Web/API/indexeddb_api#Specifications once this is merged, but it will be easy to fix so it won't break in the future.

UziTech:
Perhaps instead of changing `IndexDB 2` to `IndexDB Draft` we should just add `IndexDB Draft` and update https://developer.mozilla.org/en-US/docs/Web/API/indexeddb_api#Specifications so nothing will break.

UziTech:
@wbamberg I changed it so this PR just adds "IndexedDB Draft". We can update https://developer.mozilla.org/en-US/docs/Web/API/indexeddb_api#Specifications and remove "IndexedDB 2" after updating MDN so it won't break in between.

UziTech:
@wbamberg can this get reviewed?

wbamberg:
Honestly, as I said above, I don't think I can review this PR in the absence of a policy for what we are supposed to list in SpecData. I'm sorry to be unhelpful.

chrisdavidmills:
Apologies — I completely missed my mention in this thread somehow.

So, the trouble here is that even with this change, you'd then have to go through A LOT of IndexedDB documentation pages and manually change their spec tables.

When we had our discussion on spec tables, we basically agreed to make some changes to the structure, and then to automatically generate spec links from spec urls found in our browser compat data repo. This will be done in the future at some point, possibly in time for the next generation MDN that we are working on.

Unfortunately, the conversation didn't get to writing a policy for what specs we should include. I think that linking to the latest draft where possible would be a good idea - as you say, when the version changes this would mean the link is still correct.

An idle thought for this PR in particular. If we changed the existing IDB entries to the following, the links would all be updated automatically with correct labels. It would just mean that the data in SpecData.json is a bit screwy-looking.

```
"IndexedDB": {
    "name": "Indexed Database API 2.0",
    "url": "https://www.w3.org/TR/IndexedDB/",
    "status": "REC"
  },
  "IndexedDB 2": {
    "name": "Indexed Database API Draft",
    "url": "https://w3c.github.io/IndexedDB/",
    "status": "Draft"
  },
```

UziTech:
@chrisdavidmills I updated the PR to just change the names.

PR #1124, opened 2019-04-18
Translate {{HTTPSidebar}} to Korean

cs09g:


ExE-Boss:
The pull request name should be:
**feat: Translate {{HTTPSidebar}} to Korean**

cs09g:
@ExE-Boss Where can I find the rule for commit or PR name? so I could follow.

ExE-Boss:
I tend to use a modified [angular.js commit format](https://github.com/angular/angular.js/blob/master/DEVELOPERS.md#-git-commit-guidelines) (I capitalise the subject)

As for the subject itself, that should be a descriptive summary. For this, that would be: **“Translate {{HTTPSidebar}} to Korean”**.

wbamberg:
Also, @cs09g :

> @ExE-Boss Where can I find the rule for commit or PR name? so I could follow.

The KumaScript repository doesn't have a policy for commit messages or PR names, so don't worry about this. It is a good idea for names to be clear of course, but there are no specific rules to follow here.

PR #1125, opened 2019-04-18
chore: Fix package when installed using pnpm

ExE-Boss:
[pnpm](https://pnpm.js.org) is the third largest node package manager, which uses symlinks, hardlinks and isolated trees, which prevents packages from using packages which they haven’t declared in their dependencies.

Split from #1057

See also mdn/mdn#69

---

- `jsdom` is used extensively in tests.
- `extend` is used in **Compat.test.js**:
  https://github.com/mdn/kumascript/blob/850cf4530c3b9414c1cd4b668607b3a03f8ae30b/tests/macros/Compat.test.js#L14-L15
- ~~`jest-environment-jsdom` needs to be available from top-level `node_modules` otherwise `jest` complains that it can't find it.~~ Turns out this bug got fixed since I opened #1057.


davidflanagan:
@peterbe: I can't review this PR because I don't understand it.

@ExE-Boss: please explain why these changes to package.json are required. Is something failing? Or is the issue simply that there are problems for pnpm users that npm users don't experienced?

ExE-Boss:
The issue is that those three packages are used by the test suite, but as pnpm doesn't allow access to packages which aren't declared in `dependencies`, `devDependencies`, `optionalDependencies` or `peerDependencies` by using symlinks, hardlinks and isolated trees, the tests fail when `node_modules` was installed using pnpm.

davidflanagan:
So pnpm is working here as a linter for us catching things that might cause problems in the future.

I see that jsdom and extend are directly used by our code, and agree that they need to be added to package.json. I'm not convinced about jest-environment-jsdom. That is not anything that we're depending on directly. If it is a dependency of jest, then we shouldn't have to list it in our package.json. Could this be a bug in pnpm that that is being reported?

ExE-Boss:
needinfo?(@zkochan):
> I'm not convinced about jest-environment-jsdom. That is not anything that we're depending on directly. If it is a dependency of jest, then we shouldn't have to list it in our package.json. Could this be a bug in pnpm that that is being reported?

peterbe:
> So pnpm is working here as a linter for us catching things that might cause problems in the future.

Bothers me that we had to use `pnpm` to find that out :)
Are there other good tools on `npm` that could have caught that? If so, it would be cool to have a thing of doing that in CI. 

Also, why don't we have exact versions in the `package.json`? 
Perhaps it's moot when you have a `npm-shrinkwrap.json` because with that you get reproducible exactly the same versions anyway. 

Also, why do we have `npm-shrinkwrap.json` and not `package-lock.json`? Is that just an historical thing?

ExE-Boss:
`npm‑shrinkwrap.json` is respected during `npm install` recursively, whereas `package‑lock.json` is never published to the npm registry.

peterbe:
Answering my own question...

> Are there other good tools on npm that could have caught that? If so, it would be cool to have a thing of doing that in CI.

https://github.com/mdn/kumascript/issues/1128

peterbe:
> npm‑shrinkwrap.json is respected during npm install recursively, whereas package‑lock.json is never published to the npm registry.

We don't have a `package-lock.json` and as far as I understand, they both serve the same purpose which is also incidentally the same as `yarn.lock` if you use `yarn`. 

I remember it was `shrinkwrap` was a popular third-party package that NPM corp liked so much that they implemented it into the core `npm` program and went they did that they opted for a different name for the `.json` file.

Issue #1043, opened 2019-01-18
Requesting code freeze of GroupData and InterfaceData


ExE-Boss:
GroupData and InterfaceData are being migrated to&nbsp;<code>mdn&#x2011;data</code> in&nbsp;https://github.com/mdn/data/pull/342 and&nbsp;https://github.com/mdn/kumascript/pull/1041, and&nbsp;I&nbsp;don’t wish to&nbsp;have to&nbsp;keep updating https://github.com/mdn/data/pull/342 every&nbsp;time a&nbsp;PR&nbsp;like&nbsp;https://github.com/mdn/kumascript/pull/988 gets&nbsp;merged.

jwhitlock:
It is too early for a code freeze. I think we're at least a month away from migrating ``GroupData`` and ``InterfaceData`` to ``mdn-data``, and there are several steps to take first. For example, we need to move to a local data model to load this data, similar to BCD data.

ExE-Boss:
Also there’s https://github.com/mdn/data/issues/240#issuecomment-418479029.

PR #1126, opened 2019-04-18
refactor(macros/CSS): Use `mdn‑data` as a node package

ExE-Boss:
As requested by @jwhitlock in https://github.com/mdn/kumascript/issues/1043#issuecomment-458689142.

This will prevent changes to or outages of the GitHub API from causing CSS macros to fail.

review?(@davidflanagan, @wbamberg)

---

Note that this will require `mdn‑data` to switch to a weekly release model like the one used by BCD.


package.json
```diff
@@ -27,6 +27,7 @@
         "express": "4.16",
         "lru-cache": "5.1.1",
         "mdn-browser-compat-data": "0.x",
+        "mdn-data": "2.x",
```
peterbe: Years of having used Node and stuff and I'm still learning! I had no idea you can use an `x` like that. Thanks! I love it.

By the way, how is that different from `~`? I.e. `"~2.0"` which would match, for example, "2.0.2", "2.0.3", "2.1.0", "2.9", etc.

ExE-Boss: It’s that you can use `1.2.x` to specify `>=1.2 <1.3`, which means the same thing as `~1.2`:

https://docs.npmjs.com/misc/semver#x-ranges-12x-1x-12-

Also, `~2.0` means `>=2.0 <2.1`, so it would only match `2.0.2` and `2.0.3`, but not `2.1` or `2.9`.

To get what you mean you’d use `~2` which means the same thing as `2.x`.

macros/CSSSyntax.ejs
```diff
@@ -43,13 +37,8 @@ let externallyDescribedTypesData = {
...
 };
 let data = {
-    properties: await MDN.fetchJSONResource(propertiesUrl),
-    atRules: await MDN.fetchJSONResource(atRulesUrl),
-    selectors: await MDN.fetchJSONResource(selectorsUrl),
-    types: await MDN.fetchJSONResource(typesUrl),
-    syntaxes: await MDN.fetchJSONResource(syntaxesUrl),
-    units: await MDN.fetchJSONResource(unitsUrl),
-    externallyDescribedTypes: externallyDescribedTypesData
+    ...require('mdn-data/css'),
```
davidflanagan: I'd prefer to see a line up at the top `const cssdata = require('mdn-data/css')` and then here just do `...cssdata`.  This code is technically correct, but could be confusing, and it ends up hiding a require in the middle of the file rather than putting it up top.

tests/macros/CSSxRef.test.js
```diff
@@ -177,7 +177,7 @@ const TEST_CASE = [
 // Test runner
 // ----------------------------------------------------------------------------
 
-describeMacro('cssxref', () => {
+describeMacro('CSSxRef', () => {
```
davidflanagan: I don't have an opinion about the capitalization of macro names, except that I don't think it is worth spending any effort to do it. But I do think that if the macro (and test) filenames are going to change it should be in a PR that is specific to that purpose, and not in this PR.

npm-shrinkwrap.json
```diff
@@ -1148,9 +1148,9 @@
             }
         },
         "commander": {
-            "version": "2.17.1",
-            "resolved": "https://registry.npmjs.org/commander/-/commander-2.17.1.tgz",
-            "integrity": "sha512-wPMUt6FnH2yzG95SA6mzjQOEKUU3aLaDEmzs1ti+1E9h+CsrZghRlqEM/EJ4KscsQVG8uNN4uVreUeT8+drlgg==",
+            "version": "2.20.0",
```
davidflanagan: Ugh. I hate this file. Did you update our dependencies or did you just regenerate this file and discover that it was out of date.

Keeping dependencies up to date is important, but I'd rather not do it in the same PR as this change.  If you just made that one change to package.json and followed the steps to regenerate this file, then I suppose it is fine.

ExE-Boss: This is partially a result of #1127 (Resolve high severity `npm audit` vulnerabilities) and partially a result of running `npm install` with a newer npm release.

package.json
peterbe: To my eyes `2.x` looks a more obvious/user-friendly than `~2`. 

To me, the ~ usually means home directory or "approximately". When say "~5 seconds" then "4.87 sec" matches that.

ExE-Boss:
Well, the exports of `mdn-data/css` match the `data` object structure: https://github.com/mdn/data/blob/master/css/index.js, so merging it should be fine.

escattone:
@wbamberg I'll take a look.

macros/CSSInfo.ejs
```diff
@@ -11,21 +11,7 @@ let name = $0 || (slug ? slug.split("/").pop().toLowerCase() :
     "preview-wiki-content");
 let atRule = $1;
 
-let propertiesUrl = "https://raw.githubusercontent.com/mdn/data/master/css/properties.json";
-let atRulesUrl = "https://raw.githubusercontent.com/mdn/data/master/css/at-rules.json";
-let selectorsUrl = "https://raw.githubusercontent.com/mdn/data/master/css/selectors.json";
-let typesUrl = "https://raw.githubusercontent.com/mdn/data/master/css/types.json";
-let syntaxesUrl = "https://raw.githubusercontent.com/mdn/data/master/css/syntaxes.json";
-let unitsUrl = "https://raw.githubusercontent.com/mdn/data/master/css/units.json";
-let localStringsUrl = "https://raw.githubusercontent.com/mdn/data/master/l10n/css.json";
```
escattone: In my local dev environment, I tried using this macro and it failed because `localStringsUrl` is still used below:
```
let localStrings = await MDN.fetchJSONResource(localStringsUrl);
```

escattone:
FYI, forgot to add that the other macros seemed to work fine in my local dev environment.

wbamberg:
Indeed, this is a good thing to have merged. Thanks everyone for working on it!

ExE-Boss:
Note that if&nbsp;[`::part(…)`](https://developer.mozilla.org/docs/Web/CSS/::part), [`@supports`](https://developer.mozilla.org/docs/Web/CSS/@supports) or&nbsp;[any&nbsp;other feature that got&nbsp;updated since **mdn&#x2011;data**@v2.0.2 was released](https://github.com/mdn/data/compare/v2.0.2...master) has&nbsp;its&nbsp;page re&#x2011;rendered, the&nbsp;updated data will be&nbsp;lost until the&nbsp;next **mdn&#x2011;data**&nbsp;version [gets&nbsp;released to&nbsp;npm](https://npmjs.com/package/mdn-data) and&nbsp;deployed to&nbsp;production.

wbamberg:
Good point. I just released mdn-data 2.0.3, which ought to help a bit with that.

PR #1135, opened 2019-04-28
Add tests for DefaultAPISidebar

wbamberg:
This PR is part of https://github.com/mdn/sprints/issues/1314. It adds tests for the DefaultAPISidebar macro in preparation for updating the macro to handle the new way we are organizing event reference documentation.

For some more detail on how this macro works and when it is supposed to be used (AFAICT), please see https://github.com/mdn/sprints/issues/1314#issuecomment-487251021.

We test with several different "GroupData" structures, that feature different combinations of items: missing arrays, empty arrays, arrays containing one or more entry.

Note that guides may be specified in two places:

* they may be found as subpages of the page identified by the "overview" property in GroupData,
* they may also be explicitly listed under the "guides" property of GroupData. 

We test for this using various GroupData structures in combination with non-empty and empty "subpages" structures.

If the same guide appears in GroupData.guides and also as a subpage, then it gets listed twice. This looks like a bug, but for now I've chosen to keep the current behavior and test for it, rather than fix it, because I'm not sure of the best way to fix it.

One fix I have made now is: the "overview" entry in GroupData is optional. If it's not present, then the macro is not able to look in subpages for guides (because subpages is relative to the overview page). However in the current implementation, if "overview" is missing then the macro _also_ ignores GroupData.guides (https://github.com/mdn/kumascript/blob/master/macros/DefaultAPISidebar.ejs#L43-L44). This just seems wrong, so I've changed it.


tests/macros/defaultapisidebar.test.js
```diff
@@ -0,0 +1,359 @@
...
+*/
+const fs = require('fs');
+const path = require('path');
+const subpagesFixturePath = path.resolve(__dirname, 'fixtures/defaultapisidebar/subpages.json');
+const subpagesFixture = JSON.parse(fs.readFileSync(subpagesFixturePath, 'utf8'));
+const commonl10nFixturePath = path.resolve(__dirname, 'fixtures/defaultapisidebar/commonl10n.json');
+const commonl10nFixture = fs.readFileSync(commonl10nFixturePath, 'utf8');
+const commonL10nJSON = JSON.parse(commonl10nFixture);
+const groupDataFixturePath = path.resolve(__dirname, 'fixtures/defaultapisidebar/groupdata.json');
+const groupDataFixture = fs.readFileSync(groupDataFixturePath, 'utf8');
```
ExE-Boss: I’d prefer if this used `readFixture(…)` from https://github.com/mdn/kumascript/pull/1093.

tests/macros/defaultapisidebar.test.js
```diff
@@ -0,0 +1,359 @@
...
+describeMacro('DefaultAPISidebar', function() {
+
+    // Set any fixtures that don't change from one
+    // test to another
+    beforeEachMacro(function(macro) {
+        // env.slug only has to be truthy
+        macro.ctx.env.slug = 'not undefined';
+        // Mock calls to L10n-Common and GroupData
+        const originalTemplate = macro.ctx.template;
+        macro.ctx.template = jest.fn( async (name, ...args) => {
```
ExE-Boss: I’d prefer if this used `macro.mockTemplate(…)` from https://github.com/mdn/kumascript/pull/1093.

wbamberg:
I've renamed the test script to match the macro (I assume that's what you meant). Makes sense.

I can't use `mockTemplate` because it's not been merged and I don't want this PR to get a dependency on another unmerged PR. Happy to fix it in a follow-up though.

tests/macros/fixtures/defaultapisidebar/commonl10n.json
```diff
@@ -0,0 +1,75 @@
+{
```
escattone: Just a small point that it looks like only 7 of the 12 keys are actually used by the `DefaultAPISidebar` macro (`Methods`, `Properties`, `Events`, `Interfaces`, `Guides`, `[Translate]`, and `TranslationCTA`), so this file could be smaller if desired.

tests/macros/DefaultAPISidebar.test.js
```diff
@@ -0,0 +1,359 @@
...
+    lintHTML
+} = require('./utils');
+
+/**
+* Load all the fixtures.
+*/
+const fs = require('fs');
+const path = require('path');
+const subpagesFixturePath = path.resolve(__dirname, 'fixtures/defaultapisidebar/subpages.json');
+const subpagesFixture = JSON.parse(fs.readFileSync(subpagesFixturePath, 'utf8'));
```
escattone: I think `subpagesJSON` rather than `subpagesFixture` would be a clearer name and more consistent with your naming scheme?

tests/macros/DefaultAPISidebar.test.js
```diff
@@ -0,0 +1,359 @@
...
+    // Test sublists
+    const details = dom.querySelectorAll('ol>li>details');
+    expect(details.length).toEqual(Object.keys(config.expected.details).length);
+    // "_" prefix is to suppress lint's no-unused-vars
+    let _next = 0;
+    _next = checkSubList('Guides', config, details, checkGuideItem, _next);
+    _next = checkSubList('Interfaces', config, details, checkItem, _next);
+    _next = checkSubList('Properties', config, details, checkItem, _next);
+    _next = checkSubList('Methods', config, details, checkItem, _next);
+    _next = checkSubList('Events', config, details, checkItem, _next);
```
escattone: Super nit: The `_` could be removed if this last variable assignment wasn't done, so this would work as well:
```
let next = 0;
next = checkSubList('Guides', config, details, checkGuideItem, next);
next = checkSubList('Interfaces', config, details, checkItem, next);
next = checkSubList('Properties', config, details, checkItem, next);
next = checkSubList('Methods', config, details, checkItem, next);
checkSubList('Events', config, details, checkItem, next);
```
but that's a personal style preference.

wbamberg:
Thanks for the review @escattone . All your suggestions seem like good improvements 👍

PR #1141, opened 2019-05-02
feat: Update {{APIListAlpha}} to cope with newer title formats

ExE-Boss:
I noticed that `{{APIListAlpha}}` mentioned the “Web/API/Reference” page, which has since had its content migrated to “[Web/API](https://developer.mozilla.org/docs/Web/API#Specifications)”.

Recommended to review with whitespace changes¹ hidden: [#1141?w=1](https://github.com/mdn/kumascript/pull/1141/files?w=1).

¹ This is because **VS Code** automatically trims trailing whitespace.


wbamberg:
It looks like `isInterface()` isn't called anywhere, why not just delete it?

We might still want to have the comment under `buildTitle()` which is actually called and duplicates the same logic.

Though I would also consider just deleting Web/API/Reference at this point - it's been a redirect for 4 years.

Also https://github.com/mdn/kumascript/blob/0303dbdc98b89a5e1d92bc433e9d6bd0cd8c8b85/macros/APIListAlpha.ejs#L49-L56 is special-casing for Firefox OS that can now be deleted.

ExE-Boss:
Actually, Web/API/Reference used to contain `{{APIListData}}` and not `{{ListGroups}}`, that one is actually at [Web/Guide/API](https://developer.mozilla.org/en-US/docs/Web/Guide/API).

wbamberg:
I'm honestly puzzled by the direction this PR has taken. The first version was a comment (and whitespace)-only change, totally simple and safe. I reviewed it and suggested you could actually remove the whole function as it's not being called - again, a totally simple and safe change.

The next version is a complete rewrite of the macro, maybe with some functional changes, maybe not - it's very hard to tell - plus some very minimal test code. It's again hard for me to tell whether the test code is adequate to catch any regressions because the code it's testing has also changed.

There is no description of the changes this PR makes, why they are going to be beneficial, and why you've chosen to make this set of changes all together.

You currently have 33 open PRs against this repo. I appreciate that this is frustrating for you and that we're not good at reviewing PRs quickly, but there are ways you can make it more likely that your PRs will get reviewed and merged.

* File PRs that do one clear thing. Small PRs are better than big PRs: they're easier to review and reason about. So try to avoid making extra changes that aren't related to the PR's main purpose. If a PR adds tests, just add tests, don't update the code you are testing unless you've found a clear bug with a straightforward fix, and then explain clearly what the bug is and why you've chosen to include the fix in the PR.

* Communication: in the PR description explain clearly what the PR is doing and why we should do it. For changes that aren't obviously noncontroversial, it's a very good idea to discuss them first, for example in the Discourse forum or an ADR, so we can get consensus that this is the right way to go and so people have context for PRs. If there is some particular convention you think we should use, or tool you think we should support, for example, it's a very good idea to talk about it first.

* Consistency: if a PR starts off being about one thing, it's very confusing for reviewers when it changes greatly in scope over the course of a review. For example, if a PR starts as one small change, then after some review many unrelated changes are added, this makes the reviewer reevaluate the entire PR and try to understand what its purpose is, and that makes it less likely the PR will be merged.

What makes it most likely that a PR will get attention is if it clearly aligns with our project goals. We are generally very bad at communicating our project goals in public, and that's an area we should try to improve.

ExE-Boss:
[This page](https://developer.mozilla.org/en-US/docs/User:ExE-Boss/Macro_issues/APIListAlpha) demonstrates all the issues that are fixed by this PR.

PR #1146, opened 2019-05-06
feat: Update {{APIListAlpha}} to cope with newer title formats

ExE-Boss:
The way `{{APIListAlpha}}` is currently implemented won’t be compatible with some of the planned changes to page titles, some of which are already being done [on CSS pages](https://discourse.mozilla.org/t/38861).

[This page](https://developer.mozilla.org/docs/User:ExE-Boss/Macro_issues/APIListAlpha) demonstrates what happens when any of the newer title formats are used.

## Relevant Discourse threads:
- https://discourse.mozilla.org/t/results-of-first-css-reference-title-a-b-test/38861
- https://discourse.mozilla.org/t/api-documentation-structure-idea/31967
- https://discourse.mozilla.org/t/incorrect-titles-for-method-property-articles/31641
- https://discourse.mozilla.org/t/brainstorming-updating-titles-of-css-reference-pages/23126
- https://discourse.mozilla.org/t/proposal-changing-the-titles-of-html-element-reference-pages/19112

---

review?(@a2sheppy, @escattone, @wbamberg)

### DIFF
--- a/macros/APIListAlpha.ejs
+++ b/macros/APIListAlpha.ejs
@@ -1,45 +1,65 @@
 <%
-// Creates a list of API Interfaces sorted alphabeticlly
+// Creates a list of API Interfaces sorted alphabetically
 
-var APIHref = $0 || '/en-US/docs/Web/API';
+const APIHref = $0 || '/en-US/docs/Web/API';
 
-function isInterface(pageSlug) {
-    var spaceIndex = pageSlug.indexOf('_');
-    
+/**
+ * Checks if the page is an interface page.
+ *
+ * @param {string} pageSlug
+ * @param {string[]} [tags]
+ */
+function isInterface(pageSlug, tags = null) {
+    const spaceIndex = pageSlug.indexOf('_');
+
     // If there are no spaces (represented by underscores), then we can
     // just return true, since this is by our rules an interface in that
-    // situation. Unless the slug is simply "Reference" (why the exception?).
-    
+    // situation.
+
+    // Unless the slug is simply "Reference", which is because that page
+    // used to host this macro, until it was moved to "Web/API".
+
     if ((spaceIndex == -1) && (pageSlug != "Reference")) {
         return true;
     }
-    
+
     // If the character after the space is '(', then this is assumed to be an
     // interface with a qualifier after its name (such as "(Firefox OS)").
     // These are allowed to appear in the index.
-    
+
     if ((spaceIndex < pageSlug.length-1) && (pageSlug[spaceIndex+1] == '(')) {
         return true;
     }
-    
-    // If we get to here, then it's not an interface.
-    
-    return false;
+
+    // If we get to here, then it's probably not an interface.
+
+    return containsTag(tags, 'Interface');
 }
 
-// Prepares the title of the link. This includes wrapping the appropriate
-// portion of the title with <code>. Returns null if the page should not
-// be included in the index.
-
-function buildTitle(base) {
+/**
+ * Prepares the title of the link. This includes wrapping the appropriate
+ * portion of the title with `<code>`.
+ *
+ * @param {string} base The page title
+ * @param {string} slug The page slug without the parent
+ *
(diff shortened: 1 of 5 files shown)


macros/APIListAlpha.ejs
```diff
@@ -65,74 +103,91 @@ function containsTag(tagList, tag) {
...
-                        html += "<span>" + curLetter + "</span><ul>";
+                        html += "\t<span>" + curLetter + "</span>\n\t<ul>";
                         insertedHeading = true;
                     }
-                    html += '\n<li><span class="indexListRow"><span class="indexListTerm"><a href="' + url.replace("en-US", locale) + '" title="' + wiki.escapeQuotes(summary) + '">' + builtTitle + '</a></span>' + badge + '</span></li>';
+                    html += '\n\t\t<li>' +
+                        '<span class="indexListRow">' +
+                            '<span class="indexListTerm">' +
+                                '<a href="' + url.replace("en-US", locale) + '"' +
+                                    (summary ? ` title="${mdn.escapeQuotes(summary)}"` : '') +
```
ExE-Boss: This ensures that pages with an empty `summary` won’t have a pointless empty `title` attribute, which confuses some browsers (looking at you, **IE**/**Edge**).

macros/APIListAlpha.ejs
```diff
@@ -1,45 +1,65 @@
...
-    
+
     // If there are no spaces in the page's title, the entire base
     // string is the title, so we just wrap the whole thing in <code> and
     // return it.
-    
-    if ((spaceIndex == -1) && (base != "Reference")) {
-        return "<code>" + base + "</code>";
+
+    if (spaceIndex == -1) {
```
ExE-Boss: We no longer have to check if the page title is `Reference` here, as we already determine that in the `isInterface` function using the slug, which is more reliable.

macros/APIListAlpha.ejs
```diff
@@ -48,12 +68,30 @@ function buildTitle(base) {
...
+            `<code>${
+                mdn.escapeHTML(base.substring(0, spaceIndex))
+            }</code>${
+                mdn.escapeHTML(base.substring(spaceIndex))
+            }`,
+            base,
+        ];
+    }
+
+    const match = /^(?:API )?([^ ]+)(?: (?:-|Interface))?/i.exec(mdn.escapeHTML(base));
```
ExE-Boss: This ensures that modern page names won’t break:
- https://discourse.mozilla.org/t/incorrect-titles-for-method-property-articles/31641
- https://discourse.mozilla.org/t/api-documentation-structure-idea/31967
- https://discourse.mozilla.org/t/brainstorming-updating-titles-of-css-reference-pages/23126
- https://discourse.mozilla.org/t/proposal-changing-the-titles-of-html-element-reference-pages/19112

a2sheppy: IMO this should instead be using the slug to generate the name of the CSS item, rather than trying to strip it from the title.

ExE-Boss: This is about APIs, not CSS.

a2sheppy: The fact remains that trying to strip the name of the thing out of the title is not going to be as reliable as getting it from the slug.

a2sheppy: Which this is. Dammit, never mind.

src/api/mdn.js
```diff
@@ -115,9 +115,10 @@ module.exports = {
         return string.replace(/\$\w+\$/g, replacePlaceholder);
     },
 
-    /**
-     * Given a string, escapes all quotes within it.
-     */
+    /** Escape the given string for HTML inclusion. */
+    escapeHTML: util.htmlEscape,
```
ExE-Boss: Exposing this on the `mdn` namespace is necessary to allow fixing cases where the page title contains `<` or `>`, which confuses the Kuma sanitiser.

PR #1149, opened 2019-05-10
Clean GroupData

wbamberg:
This PR removes some archived and unused groups from GroupData.json, as discussed in https://discourse.mozilla.org/t/proposal-clean-groupdata/39610.

Specifically, it removes:

Alarm API
Application Compatibility Layer
Apps
Archive API
Bluetooth API (Firefox OS)
Camera API
Data Store API
Directory Upload API
Download API
Firefox OS
HTML Microdata API
HTML Undo Manager API
Identity
Idle API
Input Port API
Inter-App Connection API
Kill Switch API
L10N API
Mozilla Payment API
MSISDN Verification API
Network Stats API
NFC API
Permissions API (Firefox OS)
Power Management API
Request Sync API
Resource Statistics API
Social API
System Update API
TV API
UDP Socket API
Voicemail API
Wake Lock API
Web Activities
WiFi Tethering API

I believe that all these groups are either archived or don't have any docs on MDN.

This list is mostly the same as in the Discourse post above, with a few differences:

* it obviously excludes all the groups that were removed in https://github.com/mdn/kumascript/pull/1136
* it excludes the "Battery API", per @a2sheppy 's feedback in that thread we'll keep that group
* it adds the "Wifi Tethering API" which also seems to be a Mozilla-specific API with no docs on MDN.

I've tested this locally using the ListGroups.ejs call in http://localhost:8000/en-US/docs/Web/API, and it looks as expected.

### DIFF
--- a/macros/GroupData.json
+++ b/macros/GroupData.json
@@ -1,42 +1,11 @@
 [
     {
-        "Alarm API": {
-            "overview":   [ "Alarm API"],
-            "interfaces": [ "MozAlarmsManager" ],
-            "methods":    [ "Navigator.mozSetMessageHandler()"],
-            "properties": [ "Navigator.mozAlarms" ],
-            "events":     []
-        },
         "Ambient Light Events": {
             "overview":   [ "Ambient Light Events"],
             "interfaces": [ "DeviceLightEvent" ],
             "methods":    [],
             "properties": [ "Window.ondevicelight" ],
             "events":     [ "Window: devicelight" ]
-        },
-        "Application Compatibility Layer": {
-            "interfaces": [ "HTMLExtAppElement"],
-            "methods":    [],
-            "properties": [],
-            "events":     []
-        },
-        "Apps": {
-            "guides":     [ { "url":   "/Apps/Build/JavaScript_API",
-                              "title": "App Installation and Management APIs" } ],
-            "interfaces": [ "DOMApplication",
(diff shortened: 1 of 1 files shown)


wbamberg:
> Because many of these groups are used from APIRef or DefaultAPISidebar macro invocations.

Can you point me to the pages that are using these groups? As far as I can tell these APIs either have no docs on MDN or have archived docs, which should not contain any macros. I did remove macros from a few archived APIs in preparation for this PR.

It's quite possible that I have missed some, which is why I started the Discourse thread. But the only candidate for rescuing that emerged from that thread was the Battery API, as noted above.

ExE-Boss:
Well, the&nbsp;“Browser&nbsp;API” group is&nbsp;being used from [Mozilla/Gecko/Chrome/API/Browser_API](https://developer.mozilla.org/en-US/docs/Mozilla/Gecko/Chrome/API/Browser_API) and relevant pages.

wbamberg:
I don't think it is: that page has has macro calls replaced by the HTML they generate, per https://discourse.mozilla.org/t/next-steps-for-kumascript/32227.

wbamberg:
@chrisdavidmills , would you mind taking a look at this one please?

chrisdavidmills:
All looks good to me, thanks @wbamberg .

Most of these are old FxOS APIs, or bits that are now contained inside other APIs.

PR #1152, opened 2019-05-15
quick fix to repair broken devtools sidebar

chrisdavidmills:
As per https://github.com/mdn/sprints/issues/1536

Can we get a quick review of this? The devtools sidebar ended up broken after the last change to it. I feel responsible, as I was the reviewer on that work.


wbamberg:
This would have been caught if we'd had a simple test to check the HTML, using the utility Ryan added that integrates the W3C validator. All our sidebars should have this.

How urgently do you want this fixed? If you can wait a couple of hours I can add such a test to this PR.

chrisdavidmills:
Thanks for the test Will, this is really useful.

I have made all the changes you asked for, plus:

* I changed the casing of the text map items so at least they follow the correct MDN casing and are consistent with each other, even if they are inconsistent with some of the other items in the toolbar. It's generally a bit of a mess really.
* I  fixed one of the slugs as I noticed it was incorrect.

chrisdavidmills:
Cool, thanks @wbamberg . And much appreciation for jumping on this so quickly.

PR #1154, opened 2019-05-15
refactor(GroupData): Fold “CSSOM View” into “CSS Object Model”

ExE-Boss:
See [this proposal on Discourse](https://discourse.mozilla.org/t/40210/18) for details.

~~I’ve also added interfaces that were added to the specs since CSSOM Group Data was last updated.~~ (Moved to: #1156)

---

**P.S.:** [`CaretPosition`](https://developer.mozilla.org/docs/Web/API/CaretPosition) is actually [defined in CSSOM View](https://drafts.csswg.org/cssom-view/#the-caretposition-interface) and not in CSSOM.

review?(@wbamberg)


wbamberg:
Thanks for the PR!

> I’ve also added interfaces that were added to the specs since CSSOM Group Data was last updated.

Please, don't do this. If the PR is to fold CSSOM View into CSSOM, let it just do that.

macros/GroupData.json
```diff
@@ -208,26 +209,22 @@
                             "CSSStyleSheet",
                             "CSSSupportsRule",
                             "LinkStyle",
+                            "MediaQueryList",
+                            "MediaQueryListEvent",
```
ExE-Boss: I kept [`MediaQueryListEvent`](https://developer.mozilla.org/docs/Web/API/MediaQueryListEvent) as it’s part of “CSSOM View”.

ExE-Boss:
review?(@wbamberg): I’ve now extracted the added interfaces into #1156.

PR #1156, opened 2019-05-16
feat(GroupData): Update “CSS Object Model” interfaces

ExE-Boss:
This&nbsp;adds interfaces that were added to&nbsp;the&nbsp;specs since CSSOM&nbsp;Group&nbsp;Data was&nbsp;last&nbsp;updated.

## Depends on:
- [x] ~~#1154~~ (merged)

---

review?(@wbamberg)


wbamberg:
Thanks for this, but I (personally) don't see the point in adding interfaces that aren't documented yet. This data is used only (AFAIK) for building sidebars, so the only real effect of adding these will be to add broken links to the sidebars.

Wouldn't it be better to update this when we document these interfaces?

PR #1168, opened 2019-05-23
Make ListGroups omit groups that don't contain an overview property

wbamberg:
This PR updates the ListGroups.ejs macro so it doesn't show groups that don't contain an `"overview"` property.

This change was previously discussed in https://discourse.mozilla.org/t/defaultapisidebar-apiref-and-groupdata/40210/18, in particular this bit:

> Another thing we could do is this. At the moment, ListGroups will use the group’s overview property to build a link, if the overview property exists. But if the overview property does not exist, ListGroups will use the group’s name to build a link (https://github.com/mdn/kumascript/blob/master/macros/ListGroups.ejs#L45-L52). This means that an author can specify the group explicitly using overview or implicitly using the group name. I think we should stop doing this, and only use overview, and update GroupData to set overview for all groups that actually have overview pages.
>
> This means ListGroups would stop showing broken links for nonexistent overview pages.
>
> Also note that this ListGroups behavior is inconsistent with APIRef and DefaultAPISidebar, which only use the overview property to get an overview URL. So making this change would make all three macros consistent, and remove a bit of obscure magic from the ListGroups macro.

The effect of this change is that:

* currently,  for all APIs that don't have overview pages (see https://github.com/mdn/sprints/issues/1537), https://developer.mozilla.org/en-US/docs/Web/API#Specifications will contain broken links
* after this change, these broken links won't be included and the page will look a little less sad.

*****

To test this change:

1. get your local Kuma running, and point it at this branch
2. scrape the Web/API page:

```
docker exec kuma_worker_1 python ./manage.py scrape_document https://developer.mozilla.org/en-US/docs/Web/API
```
3. refresh your KS with `docker-compose restart`
4. shift-refesh http://localhost:8000/en-US/docs/Web/API and observe that the list no longer includes any of the APIs listed in https://github.com/mdn/sprints/issues/1537

### DIFF
--- a/macros/ListGroups.ejs
+++ b/macros/ListGroups.ejs
@@ -42,13 +42,12 @@ for(let name of groupNames) {
   let groupObj = groupData[name];
   let firstLetter = name[0];
   let groupOutput = '';
-  let overviewName = name;
   let badge = "";
 
-  if (groupObj.hasOwnProperty("overview")) {
-    overviewName = groupObj.overview;
+  if (groupObj.overview === undefined) {
+    continue;
   }
-
+  let overviewName = groupObj.overview;
   let groupUrl = APIHref + "/" + spacesToUnderscores(htmlEscape(overviewName));
   let page = await wiki.getPage(groupUrl);
 
--- a/tests/macros/ListGroups.test.js
+++ b/tests/macros/ListGroups.test.js
@@ -23,7 +23,7 @@ const groupDataFixture = fs.readFileSync(groupDataFixturePath, 'utf8');
  */
 const overviewPages = {
     '/en-US/docs/Web/API/An_overview_page_for_ATestInterface_API': { tags: ['foo', 'bar'] },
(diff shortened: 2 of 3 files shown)


wbamberg:
This has now gone out, and it's looking better: https://developer.mozilla.org/en-US/docs/Web/API#Specifications although I'm very annoyed to have missed Resize Observer, which managed to deceive me by having an "overview" property pointing to a nonexistent page. Sigh.

wbamberg:
> This has now gone out, and it's looking better: https://developer.mozilla.org/en-US/docs/Web/API#Specifications although I'm very annoyed to have missed Resize Observer, which managed to deceive me by having an "overview" property pointing to a nonexistent page. Sigh.

Thanks to Ryan for merging https://github.com/mdn/kumascript/pull/1170 and doing another push, this has now gone, and all the links in https://developer.mozilla.org/en-US/docs/Web/API#Specifications are blue \o/.

PR #1190, opened 2019-06-23
Add entry for CSSPseudoElement to InterfaceData macro

lpd-au:
`CSSPseudoElement` [inherits from `EventTarget`](https://www.w3.org/TR/css-pseudo-4/#ref-for-eventtarget) and [includes `Animatable`](https://www.w3.org/TR/web-animations-1/#extensions-to-the-pseudoelement-interface).

MDN page: https://developer.mozilla.org/en-US/docs/Web/API/CSSPseudoElement

### DIFF
--- a/macros/InterfaceData.json
+++ b/macros/InterfaceData.json
@@ -185,6 +185,12 @@
       "inh": "CSSValue",
       "impl": []
     },
+    "CSSPseudoElement": {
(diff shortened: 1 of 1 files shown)


lpd-au:
@chrisdavidmills you may be a good reviewer for this

PR #1203, opened 2019-07-15
Added XSLT specifications to SpecData macro

SebastianZ:
As the title says, this adds the three XSLT specifications released so far. See https://www.w3.org/TR/?title=xsl%20transformations.

There's actually a fourth one called ["XSL Transformations (XSLT) Version 2.0 (Second Edition)"](https://www.w3.org/TR/2009/PER-xslt20-20090421/), but it looks like it never reached the recommendation status.

Also, I used the short names for those specs like it's done for other specifications. "XSLT 3.0" is just easier recognizible than "XSL Transformations (XSLT) Version 3.0".

Sebastian

### DIFF
--- a/macros/SpecData.json
+++ b/macros/SpecData.json
@@ -1664,16 +1664,6 @@
     "url": "https://www.w3.org/TR/xpath-31/",
     "status": "REC"
   },
-  "XSLT 1.0": {
-    "name": "XSLT 1.0",
-    "url": "https://www.w3.org/TR/xslt-10/",
-    "status": "REC"
-  },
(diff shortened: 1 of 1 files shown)


SebastianZ:
I see. I didn't know about that decision.

I have to admit that I also just started to learn about XSLT, but it seems that the newer spec. versions supersede the older ones. Only XSLT 2.0 had [some incompatible changes](https://www.w3.org/TR/xslt20/#changes) to version 1.0.

But I'm fine with only adding 3.0 for now. If the need for an older version comes up at some point, we can still add it later.

Sebastian

PR #1207, opened 2019-07-29
adding overview page to console api groupdata

chrisdavidmills:
As per https://github.com/mdn/sprints/issues/1582#issuecomment-515617774

### DIFF
--- a/macros/GroupData.json
+++ b/macros/GroupData.json
@@ -96,6 +96,7 @@
                             "Element: paste" ]
         },
         "Console API": {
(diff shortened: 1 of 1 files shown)


chrisdavidmills:
Gah, I'm not really sure how to fix that, and don't really have time to start digging into tests right now. Hrm. I wonder if it is better to just leave it as uppercase right now, so we can get it mwrged, even if it is strictly speaking wrong. I think I'll do that.

PR #1209, opened 2019-07-29
Add WebXR Device API

jpmedley:
@a2sheppy and I are about to start creating content for this because both Chrome and FF are hoping to ship this enabled by default sometime in the fall.


jpmedley:
@a2sheppy FYI

jpmedley:
@a2sheppy Is there anything else I need to do?

PR #1216, opened 2019-08-18
just edit to try to remove bug

NorinMp143:
I am just trying to remove bug


NorinMp143:
there was a problem in bug.ejs file that starting < extra thing so i'll try to manage

macros/bug.ejs
```diff
@@ -77,7 +76,7 @@ var s_bug = mdn.localString({
     "pl"   : "błąd",
     "de"   : "Bug",
     "ru"   : "баг"
-});
+});
```
peterbe:
```suggestion
});
```

PR #1223, opened 2019-08-29
updating resize observer groupdata

chrisdavidmills:
This PR does tow things, both related to the GroupData for Resize Observer API:

1. Removes ResizeObservation form the list of interfaces. As per https://drafts.csswg.org/resize-observer/#resize-observation-interface, it is an internal object not exposes to JavaScript, so there is no point including it in MDN docs.
2. Removes the ResizeObserverCallback. We can argue about this later, but I need to get this documented quickly, and I really don't see the point in having a separate page for the callback, which is really part of the constructor and won't be used anywhere else.

### DIFF
--- a/macros/GroupData.json
+++ b/macros/GroupData.json
@@ -953,6 +953,7 @@
                             "ServiceWorkerGlobalScope: pushsubscriptionchange" ]
         },
         "Resize Observer API": {
+            "overview":   [ "Resize Observer API" ],
             "interfaces": [ "ResizeObserver",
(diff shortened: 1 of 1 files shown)


chrisdavidmills:
@Elchi3 can you give this a quick review, please sir?

chrisdavidmills:
@Elchi3 cheers mate. 

Just to let you know, I updated it slightly just then to add the overview page into groupdata. I'm planning to update it today.

PR #1127, opened 2019-04-19
fix: Resolve high severity `npm audit` vulnerabilities

ExE-Boss:
review?(@davidflanagan, @peterbe, @wbamberg)

PR #1224, opened 2019-08-29
fix: Resolve critical severity `npm audit` vulnerabilities

ExE-Boss:
review?(@a2sheppy, @davidflanagan, @Elchi3, @peterbe, @wbamberg)

---

I also kind‑of forgot about updating `jest` as a result of my old PC dying.

I’ll do that as soon as this is merged.

### DIFF
--- a/npm-shrinkwrap.json
+++ b/npm-shrinkwrap.json
@@ -50,6 +50,41 @@
                 "methods": "^1.1.2"
             }
         },
+        "@types/json-schema": {
+            "version": "7.0.3",
+            "resolved": "https://registry.npmjs.org/@types/json-schema/-/json-schema-7.0.3.tgz",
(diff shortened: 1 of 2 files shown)


peterbe:
Can you share what commands you ran to make this upgrade? `npm audit fix` (or whatever it's called, only fixes the shrinkwrap right?)

Also, can we just get rid of shrinkwrap and switch to the builtin `package-lock.json`?

ExE-Boss:
`npm audit --fix` and `npm update`

peterbe:
Is there no way to make these changes smaller?

ExE-Boss:
I can split it into two commits like in #1127, which I probably should’ve done from the get‑go.

peterbe:
> I can split it into two commits like in #1127, which I probably should’ve done from the get‑go.

That would be much appreciated! The smaller the atoms the better. 

Here's the review effect of few vs too many changes: 
source

ExE-Boss:
Moved to #1226 and #1227.

PR #1228, opened 2019-08-31
chore(package): Update Jest to version 24.9

ExE-Boss:
Just like I promised in #1224.

Depends on: #1057, as I wasn’t able to verify if the update worked otherwise.

### DIFF
--- a/npm-shrinkwrap.json
+++ b/npm-shrinkwrap.json
@@ -13,6 +13,127 @@
                 "@babel/highlight": "^7.0.0"
             }
         },
+        "@babel/core": {
+            "version": "7.5.5",
+            "resolved": "https://registry.npmjs.org/@babel/core/-/core-7.5.5.tgz",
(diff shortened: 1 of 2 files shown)


peterbe:
Now, this PR is mixing two different things. One is a jest upgrade, the other is a fix for the tests under windows. **Why can't this be two distinct PRs?**

ExE-Boss:
So, can you merge #1057 first, so that I can verify whether the update work correctly?

(it’s SemVer‑breaking, so I need to test it locally)

peterbe:
@ExE-Boss I got emails that seem to indicate that you made different git commits. But it certainly appears that you have squashed and rebased it down to 1 commit. If you need to do that, close the PR first otherwise we can't see what changed and it makes reviewing frustrating. Please. If you need to rebase because `master` has "changed under your feet" just rebase without changing any of the commits.

ExE-Boss:
I only removed the commits that were merged into master (i.e. #1057)

PR #1245, opened 2019-09-23
feat(macros): Add {{EmbedTest262ReportResultsTable($0)}} to enable embeds of Test262.report results

rwaldron:
@Elchi3 instructed me to create and submit a macro for this new feature that we're testing. We're expecting to announce it on October 1st. 

 Here's what the embed looks like: 

image

And in place: 

image

The strange gray background will be fixed and deployed by Wednesday of this week.

### DIFF
--- a/macros/EmbedTest262ReportResultsTable.ejs
+++ b/macros/EmbedTest262ReportResultsTable.ejs
@@ -2,11 +2,12 @@
 // Embeds a Test262.report results table
 //
 // Parameters:
-//  $0 - An embed URL from test262.report (absolute)
-//  Example call {{EmbedTest262ReportResultsTable("https://test262.report/embed/features/globalThis?engines=chakra%2Cjavascriptcore%2Cspidermonkey%2Cv8&summary=true&include-browsers=true")}}
+//  $0 - An embed URL from test262.report (relative)
+//
+//  Example call {{EmbedTest262ReportResultsTable("/features/globalThis?engines=chakra%2Cjavascriptcore%2Cspidermonkey%2Cv8&summary=true&include-browsers=true")}}
 //
 
-var str = '<iframe frameborder="0" scrolling="no" style="width:100%;height:300px;" src="' + $0 + '"></iframe>';
+var str = '<iframe frameborder="0" scrolling="no" style="width:100%;height:300px;" src="https://test262.report/embed' + $0 + '"></iframe>';
 
 %>
 <%- str %>
--- a/macros/EmbedTest262ReportResultsTable.ejs
+++ b/macros/EmbedTest262ReportResultsTable.ejs
@@ -2,12 +2,14 @@
 // Embeds a Test262.report results table
 //
 // Parameters:
-//  $0 - An embed URL from test262.report (relative)
+//  $0 - A "feature" name from test262.report
 //
-//  Example call {{EmbedTest262ReportResultsTable("/features/globalThis?engines=chakra%2Cjavascriptcore%2Cspidermonkey%2Cv8&summary=true&include-browsers=true")}}
+//  Example calls:
+// {{EmbedTest262ReportResultsTable("globalThis")}}
+// {{EmbedTest262ReportResultsTable("BigInt")}}
 //
 
-var str = '<iframe frameborder="0" scrolling="no" style="width:100%;height:300px;" src="https://test262.report/embed' + $0 + '"></iframe>';
+var str = '<iframe frameborder="0" scrolling="no" style="width:100%;height:300px;" src="https://test262.report/embed/features/' + $0 + '?engines=chakra%2Cjavascriptcore%2Cspidermonkey%2Cv8&summary=true&include-browsers=true"></iframe>';
 
 %>
 <%- str %>
--- a/macros/EmbedTest262ReportResultsTable.ejs
+++ b/macros/EmbedTest262ReportResultsTable.ejs
@@ -8,8 +8,5 @@
 // {{EmbedTest262ReportResultsTable("globalThis")}}
 // {{EmbedTest262ReportResultsTable("BigInt")}}
 //
-
-var str = '<iframe frameborder="0" scrolling="no" style="width:100%;height:300px;" src="https://test262.report/embed/features/' + $0 + '?engines=chakra%2Cjavascriptcore%2Cspidermonkey%2Cv8&summary=true&include-browsers=true"></iframe>';
-
 %>
-<%- str %>
+<iframe frameborder="0" scrolling="no" style="width: 100%; height: 300px;" src="https://test262.report/embed/features/<%= $0 %>?engines=chakra%2Cjavascriptcore%2Cspidermonkey%2Cv8&summary=true&include-browsers=true"></iframe>
--- a/macros/EmbedTest262ReportResultsTable.ejs
+++ b/macros/EmbedTest262ReportResultsTable.ejs
@@ -9,4 +9,4 @@
 // {{EmbedTest262ReportResultsTable("BigInt")}}
 //
 %>
-<iframe frameborder="0" scrolling="no" style="width: 100%; height: 300px;" src="https://test262.report/embed/features/<%-encodeURIComponent($0)%>?engines=chakra%2Cjavascriptcore%2Cspidermonkey%2Cv8&summary=true&include-browsers=true"></iframe>
+<iframe width="100%" height="300" src="https://test262.report/embed/features/<%-encodeURIComponent($0)%>?engines=chakra%2Cjavascriptcore%2Cspidermonkey%2Cv8&summary=true&include-browsers=true"></iframe>
--- a/macros/EmbedTest262ReportResultsTable.ejs
+++ b/macros/EmbedTest262ReportResultsTable.ejs
@@ -9,4 +9,4 @@
 // {{EmbedTest262ReportResultsTable("BigInt")}}
 //
 %>
-<iframe frameborder="0" scrolling="no" style="width: 100%; height: 300px;" src="https://test262.report/embed/features/<%= $0 %>?engines=chakra%2Cjavascriptcore%2Cspidermonkey%2Cv8&summary=true&include-browsers=true"></iframe>
+<iframe frameborder="0" scrolling="no" style="width: 100%; height: 300px;" src="https://test262.report/embed/features/<%-encodeURIComponent($0)%>?engines=chakra%2Cjavascriptcore%2Cspidermonkey%2Cv8&summary=true&include-browsers=true"></iframe>
--- a/macros/EmbedTest262ReportResultsTable.ejs
+++ b/macros/EmbedTest262ReportResultsTable.ejs
@@ -1,0 +1,12 @@
+<%
+// Embeds a Test262.report results table
+//
+// Parameters:
+//  $0 - An embed URL from test262.report (absolute)
+//  Example call {{EmbedTest262ReportResultsTable("https://test262.report/embed/features/globalThis?engines=chakra%2Cjavascriptcore%2Cspidermonkey%2Cv8&summary=true&include-browsers=true")}}
+//
+
+var str = '<iframe frameborder="0" scrolling="no" style="width:100%;height:300px;" src="' + $0 + '"></iframe>';
+
+%>
+<%- str %>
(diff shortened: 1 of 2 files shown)


macros/EmbedTest262ReportResultsTable.ejs
```diff
@@ -0,0 +1,12 @@
+<%
+// Embeds a Test262.report results table
+//
+// Parameters:
+//  $0 - An embed URL from test262.report (absolute)
```
Elchi3: The reason why https://github.com/mdn/kumascript/blob/master/macros/EmbedGHLiveSample.ejs does use relative URLs is that we don't want to ever allow people to provide arbitrary iframe URLs. Please change this to a relative parameter so that we can make sure people can only ever use test262.report URLs here.

rwaldron: That makes perfect sense. I will push a revision shortly. Thanks for the quick review!

rwaldron:
@Elchi3 new commit added!

Elchi3:
Thanks for the quick followup, Rick! I wonder if we also want to get rid of the longish URL parameters? Do you expect these things to change or do you want random MDN contributors to fiddle with it? I guess editor friendly would be if authors could just write `{{EmbedTest262ReportResultsTable("globalThis")}}`

Also, I will need to request from a Kuma/KumaScript engineer. I don't know who could review this best at the moment. @tobinmori, can you get a reviewer for this change?

I'm trying to describe the experiment briefly for context. @rwaldron please provide additional information if I forgot anything.

MDN documents ECMAScript features and web developers are interested in links to the feature's specs and they also need browser compatibility data to see if they can use the feature. MDN provides this info at the bottom of the page already. See e.g. https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Global_Objects/globalThis#Specifications

Now Bocoup has built https://test262.report. It reports implementation of ECMAScript features in browser engines directly from the ECMAScript test suite. We would like to do an experiment if MDN readers are interested in reading implementation status in browsers directly from the test262 output along with browser compat info that we already provide. For that, Bocoup has created an embeddable page, like this one for globalThis: https://test262.report/embed/features/globalThis?engines=chakra%2Cjavascriptcore%2Cspidermonkey%2Cv8&summary=true&include-browsers=true

The screenshot above shows this embed experiment in action. The embed contains "Is this helpful?" to figure out if this experiment is valuable.

For implementation of this experiment, I advised to use the same mechanism MDN uses for embedding code examples from mdn.github.io. See https://github.com/mdn/kumascript/blob/master/macros/EmbedGHLiveSample.ejs.

So, this PR is basically the same, but instead of mdn.github.io pages, it allows embedding from test262.report/embed/ so the iframe can be rendered on the page. I haven't worried about performance and we only expect this to be added to maybe 10 to 15 pages. (this only makes sense for new ECMAScript features).

Our product owner Kadir @atopal has been involved when this experiment was first presented and agreed to get something in to test this out. Bocoup agreed to do all the implementation, so here it is :)

rwaldron:
@Elchi3 

> I guess editor friendly would be if authors could just write `{{EmbedTest262ReportResultsTable("globalThis")}}`

Done!

macros/EmbedTest262ReportResultsTable.ejs
```diff
@@ -0,0 +1,15 @@
...
+// Embeds a Test262.report results table
+//
+// Parameters:
+//  $0 - A "feature" name from test262.report
+//
+//  Example calls:
+// {{EmbedTest262ReportResultsTable("globalThis")}}
+// {{EmbedTest262ReportResultsTable("BigInt")}}
+//
+
```
ExE-Boss:
```suggestion
%>
```

macros/EmbedTest262ReportResultsTable.ejs
```diff
@@ -0,0 +1,15 @@
...
+var str = '<iframe frameborder="0" scrolling="no" style="width:100%;height:300px;" src="https://test262.report/embed/features/' + $0 + '?engines=chakra%2Cjavascriptcore%2Cspidermonkey%2Cv8&summary=true&include-browsers=true"></iframe>';
```
ExE-Boss:
```suggestion
<iframe frameborder="0" scrolling="no" style="width: 100%; height: 300px;" src="https://test262.report/embed/features/<%= $0 %>?engines=chakra%2Cjavascriptcore%2Cspidermonkey%2Cv8&summary=true&include-browsers=true"></iframe>
```

macros/EmbedTest262ReportResultsTable.ejs
```diff
@@ -0,0 +1,15 @@
...
+
```
ExE-Boss:
```suggestion
```

macros/EmbedTest262ReportResultsTable.ejs
```diff
@@ -0,0 +1,15 @@
...
+%>
```
ExE-Boss:
```suggestion
```

macros/EmbedTest262ReportResultsTable.ejs
```diff
@@ -0,0 +1,15 @@
...
+<%- str %>
```
ExE-Boss:
```suggestion
```

rwaldron:
@ExE-Boss thanks for the review, the recommended changes have been made!

rwaldron:
@ExE-Boss @Elchi3 is there anything further that's needed from me?

ExE-Boss:
This should have tests written for it.

tobinmori:
Thanks @Elchi3. I'll add this to our backlog and find someone to assign it to. . .

Can you tell me how urgent getting this merged is? High/Med/Low?

Just so you know we are currently experiencing a logjam with all the code we have to review.

rwaldron:
@tobinmori thanks! We had planned with @Elchi3 and @atopal to have this deployed along with announcement in time for View Source, but unfortunately we've missed that window :|

rwaldron:
@ExE-Boss tests added!

macros/EmbedTest262ReportResultsTable.ejs
```diff
@@ -0,0 +1,12 @@
...
+//
+// Parameters:
+//  $0 - A "feature" name from test262.report
+//
+//  Example calls:
+// {{EmbedTest262ReportResultsTable("globalThis")}}
+// {{EmbedTest262ReportResultsTable("BigInt")}}
+//
+%>
+<iframe frameborder="0" scrolling="no" style="width: 100%; height: 300px;" src="https://test262.report/embed/features/<%= $0 %>?engines=chakra%2Cjavascriptcore%2Cspidermonkey%2Cv8&summary=true&include-browsers=true"></iframe>
```
peterbe: Does there need to be vetting of that `$0` parameter? I.e. if someone edits a wiki page with:
```
{{EmbedTest262ReportResultsTable('"><script>alert()</script>')}}
```
Or is that already taken care of in the kuma-to-kumascript stuff? @escattone ?

rwaldron: @peterbe HTML sanitization appears to be built in, ie. `'"><script>alert()</script>'` would become: `&#34;&gt;&lt;script&gt;alert()&lt;/script&gt;`

ExE-Boss: Actually, we need URI escaping for this, instead of HTML escaping, which is what `<%=` does:

```suggestion
<iframe
    frameborder="0" scrolling="no" style="width: 100%; height: 300px;"
    src="https://test262.report/embed/features/<%-
        encodeURIComponent($0)
    %>?engines=chakra%2Cjavascriptcore%2Cspidermonkey%2Cv8&summary=true&include-browsers=true"
></iframe>
```

rwaldron: @ExE-Boss I don't mind making this change, but I was curious to see how this is used in other macros, so I did a search and no other macros are using `encodeURIComponent`—is that a new requirement?

rwaldron:
@Elchi3 @ExE-Boss @tobinmori @peterbe @escattone recommended changes made, ready for review.

macros/EmbedTest262ReportResultsTable.ejs
```diff
@@ -0,0 +1,12 @@
...
+//
+// Parameters:
+//  $0 - A "feature" name from test262.report
+//
+//  Example calls:
+// {{EmbedTest262ReportResultsTable("globalThis")}}
+// {{EmbedTest262ReportResultsTable("BigInt")}}
+//
+%>
+<iframe frameborder="0" scrolling="no" style="width: 100%; height: 300px;" src="https://test262.report/embed/features/<%-encodeURIComponent($0)%>?engines=chakra%2Cjavascriptcore%2Cspidermonkey%2Cv8&summary=true&include-browsers=true"></iframe>
```
escattone: Currently on MDN, when bleaching the document content, we do not allow `scrolling` or `style` attributes on `iframe` elements (and I don't think we need the `frameborder` attribute?). When I tested this locally, I used:

```html
<iframe width="100%" height="300px" src="..."></iframe>
```

Also, MDN will strip-out any `iframe` sources that are not part of a whitelist (see https://github.com/mozilla/kuma/blob/8ac987263f711cb3662b01059cc8409f370c7775/kuma/settings/common.py#L1247), so I'll need to create a companion PR in Kuma to add `parse_iframe_url('https://test262.report/embed/features/')` to that whitelist in order for this to work.

escattone:
```suggestion
<iframe width="100%" height="300px" src="https://test262.report/embed/features/<%-encodeURIComponent($0)%>?engines=chakra%2Cjavascriptcore%2Cspidermonkey%2Cv8&summary=true&include-browsers=true"></iframe>
```

escattone:
> ...but also I need to create a companion PR for this in Kuma as noted below.

See https://github.com/mozilla/kuma/pull/5916

escattone:
Oh, one more thing. When I tested this locally, I noticed that there is a large whitespace gap immediately after where the macro is placed in the document:

image

Is that intended?

macros/EmbedTest262ReportResultsTable.ejs
ExE-Boss: I’d recommend adding line breaks like in https://github.com/mdn/kumascript/pull/1245#discussion_r330833257, which is what Prettier gave me.

rwaldron: Thanks for the review @escattone. I'm a little confused by some of this feedback, as it appears to contradict existing macro conventions—maybe you can help me better understand :D 

> we do not allow `scrolling` or `style` attributes

- `style`: Is this restriction limited to `iframe`?
- `scrolling`: this appears in `EmbedNewRelicChart.ejs`—perhaps that predates the restriction? I can remove it

> `<iframe width="100%" height="300px" src="..."></iframe>`

Works for me, I will make this change. 

> Also, MDN will strip-out any iframe sources that are not part of a whitelist

Good to know—this hadn't been mentioned prior to now. I can make that PR if you'd like?

Thanks again!

escattone: >     * `style`: Is this restriction limited to `iframe`?

Yes, that's correct.

>     * `scrolling`: this appears in `EmbedNewRelicChart.ejs`—perhaps that predates the restriction? I can remove it

Even though the `EmbedNewRelicChart.ejs` macro uses the `scrolling` attribute, it is removed by the bleaching process after the document is rendered, so there's nothing gained by using it in the macro.

rwaldron:
@escattone regarding the height: The page that's displayed in the iframe is responsive, so when it's viewed on desktop, its width > height, but on mobile its height > width. Without responsive styles in the host page, I think we're stuck with a hard height value (I don't see any macros that use the "iframe container" trick for dynamic iframe height). When I looked at `EmbedInteractiveExample.ejs`, I saw that it was given a class that appears to be provided by the host MDN pages.

macros/EmbedTest262ReportResultsTable.ejs
rwaldron: @ExE-Boss is this the new preferred convention? I don't see any other examples like this in the existing macros 🤷‍♀

escattone: No, it's not the preferred convention, since we don't use `prettier` for `ejs` files. The single-line approach you have is fine.

rwaldron:
@escattone when will this be available for use on MDN?

ExE-Boss:
There’s a new deployment almost every Thursday, so probably in half of a week.

escattone:
> @escattone when will this be available for use on MDN?

I'm sorry @rwaldron, I should have let you know! I deployed this later on the same day it was merged, so it has been available since last Thursday afternoon around 4:40pm Pacific time.

rwaldron:
@escattone thanks!

PR #1249, opened 2019-10-01
take Kumascript Dockerfile from kuma

escattone:
This is a companion PR to https://github.com/mozilla/kuma/pull/5902. It should be reviewed and merged along with that PR.

This PR moves the Kumascript `Dockerfile` (and associated `README.md` file) from the Kuma repo into this repo where it belongs. It also takes advantage of the new `make build-kumascript-with-all-tags` command provided by https://github.com/mozilla/kuma/pull/5902.

### DIFF
--- a/docker/Dockerfile
+++ b/docker/Dockerfile
@@ -1,0 +1,46 @@
+FROM node:10.16.3@sha256:e0f7dabe991810057aee6ba7a7d4136fe7fc70e99235d79e6c7ae0177656a5c8
+
+RUN set -ex && \
+    apt-get update && \
+    apt-get install -y --no-install-recommends \
+        gettext \
+        mime-support \
+        build-essential \
+        python2.7 \
+        default-jre \
+    && rm -rf /var/lib/apt/lists/*
+
+# remove the node user from the base package and add a non-privileged user
+RUN userdel --force --remove node && \
+    adduser --uid 1000 --disabled-password --gecos '' --no-create-home kumascript
+
+ARG REVISION_HASH
+# make the git commit hash permanently available within this image.
+ENV REVISION_HASH $REVISION_HASH
+
+WORKDIR /
+
+COPY kumascript/package.json kumascript/npm-shrinkwrap.json /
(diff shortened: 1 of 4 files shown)


escattone:
@peterbe Do I need to create/update some settings file for `renovate` so that it begins updating `docker/Dockerfile`?

escattone:
> @peterbe Do I need to create/update some settings file for `renovate` so that it begins updating `docker/Dockerfile`? Or maybe we're not even using `renovate` yet for this repo and I need to configure that as well?

Ah, yeah, it looks like we're not using `renovate` yet in this repo. I'll look into setting it up once this lands.

docker/Dockerfile
```diff
@@ -0,0 +1,46 @@
...
+# make the git commit hash permanently available within this image.
+ENV REVISION_HASH $REVISION_HASH
+
+WORKDIR /
+
+COPY kumascript/package.json kumascript/npm-shrinkwrap.json /
+RUN npm config set python /usr/bin/python2.7 && \
+    # install the Node.js dependencies,
+    # with versions specified in npm-shrinkwrap.json
+    npm ci && \
```
ExE-Boss: `npm ci` is an alias to `npm clean‑install`
```suggestion
    npm clean-install && \
```

escattone:
@peterbe I removed you as a reviewer since I realized that I was pulling your mind into this realm when you've already got so much on your plate, but of course feel free to review if you'd like to.

docker/Dockerfile
escattone: @ExE-Boss It's my understanding that `npm ci` is the proper command. In fact, I don't see any `npm` documentation on a `npm clean-install` command, and I think the `ci` references `continuous integration` rather than `clean install`. In any case, I'm going to keep this as it is, since changing it would be outside the scope of this PR anyway.

ExE-Boss: It is as of https://github.com/npm/cli/pull/57.

escattone: Oh! Good to know, thanks @ExE-Boss, but I'll leave that for another PR.

PR #1251, opened 2019-10-04
add initial config for renovate

escattone:
Proposed initial config for `renovate`. `renovate` is not yet enabled for this repo, but I'd like to enable it once #1249 lands in order to pin the digests in the Dockerfile. This config also exposes `package.json` to `renovate`, but we might not want to do that yet.

### DIFF
--- a/renovate.json
+++ b/renovate.json
@@ -1,0 +1,12 @@
+{
+  "extends": [
+    "config:base"
+  ],
+  "masterIssue": true,
+  "dockerfile": {
+    "pinDigests": true
+  },
+  "rebaseStalePrs": true,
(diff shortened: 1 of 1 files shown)


escattone:
@peterbe Good points! I don't think we're ready yet for the onslaught of Renovate PRs to update dependencies in package.json, and like you said, we'd probably have to upgrade to using `package-lock.json` anyway, so I'll add the `"ignorePaths": ["package.json"]` here, and we'll just focus on pinning the digest in the Dockerfile.

peterbe:
If it doesn't work, we can tune it later.

PR #1266, opened 2019-10-29
fix(JSRef): Add `<wbr/>` to property names

ExE-Boss:
### Before
<table><td>Screenshot_2019-10-29 – JSRef Before</td></table>

### After
<table><td>Screenshot_2019-10-29 – JSRef After</td></table>


macros/JSRef.ejs
```diff
@@ -161,8 +164,9 @@ function buildSublist(pages, title, opened) {
 
     var summary = escapeQuotes(aPage.summary) || '';
     var url = aPage.url.replace('en-US', locale);
-    var title = aPage.title;
+    let title = aPage.title;
```
ExE-Boss: The function takes a&nbsp;`title` argument, and&nbsp;due to&nbsp;the&nbsp;weird hoisting behaviour of&nbsp;`var`, it&nbsp;might&nbsp;have gotten&nbsp;overwritten, instead&nbsp;of&nbsp;just getting&nbsp;shadowed, like&nbsp;with&nbsp;`let`.

PR #1270, opened 2019-10-29
fix bogus jest/valid-describe linting error

### LINKED ISSUE #1269: lint error is blocking builds
The following `eslint` error is breaking builds:

image

It's annoying, because in our case, the `First argument must be name  jest/valid-describe` linting error is a total false alarm, and in fact, the check itself is deeply flawed (for example, see discussion at https://github.com/jest-community/eslint-plugin-jest/issues/203).

escattone:
Fixes #1269 

Upgrading to `eslint` version `6.6.0` gets rid of the bogus error, but flags a new one . The new error seemed legit, so I fixed it. NOTE: The updates to `package.json` and `npm-shrinkwrap.json` were both done via `npm install eslint@6.6.0`.

### DIFF
--- a/npm-shrinkwrap.json
+++ b/npm-shrinkwrap.json
@@ -128,9 +128,9 @@
             }
         },
         "acorn-jsx": {
-            "version": "5.0.1",
-            "resolved": "https://registry.npmjs.org/acorn-jsx/-/acorn-jsx-5.0.1.tgz",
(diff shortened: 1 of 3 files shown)


escattone:
Wow, this worked locally, but fails in Travis.

escattone:
I'm closing this in favor of a better approach.

PR #1272, opened 2019-11-01
Adding Shapes 2, Writing Modes to PR

rachelandrew:
Adding the CSS Shapes Level 2 spec: https://drafts.csswg.org/css-shapes-2

The Writing Modes spec is now a Proposed Rec: https://www.w3.org/TR/css-writing-modes-3/


ExE-Boss:
See also: https://github.com/mdn/data/pull/397

PR #1276, opened 2019-11-04
Fix to update the class to remove two bins problem in obsolete header

velusgautam:
**Obsolete** svg icon is calling a **mixin** that auto generates the **::before** bin to obsolete and obsoleteBadge only has the same red background
Screenshot 2019-11-04 at 11 08 41

Before Fix
Screenshot 2019-11-04 at 10 58 34

After fix of class

Screenshot 2019-11-04 at 11 06 12


peterbe:
CC @ExE-Boss Can you consolidate how this relates to your suggested fix?

velusgautam:
@ExE-Boss @peterbe Can I close this PR?

Issue #1274, opened 2019-11-02
Dockerfile uses Python 2.7


callahad:
The kumascript Dockerfile calls `npm config set python /usr/bin/env/python2.7` to tell `node-gyp` which Python to use when building native extensions for Node.

Python 2.7 goes end-of-life at the end of the year.

We should update that to a compatible Python 3 and make sure everything still builds and runs as expected.

callahad:
See https://github.com/nodejs/node-gyp#configuring-python-dependency for info on how node-gyp finds a Python to use

May be more obvious to just set the `PYTHON` environment variable (and add a comment in the Dockerfile as to why)

escattone:
Thanks for opening this issue @callahad! This has been on my mind for some time, and I started work on a resolution for this a while back, but was pulled into something else before I could submit a PR and then I forgot about it in the midst of everything else going on. My original thought here was that we should remove any explicit configuration of Python altogether, and just use whatever comes bundled within the `node` image, and I think that still applies. See https://github.com/mdn/kumascript/pull/1278

PR #1278, opened 2019-11-12
remove explicit python from Dockerfile

escattone:
Fixes https://github.com/mdn/kumascript/issues/1274

Do not explicitly set the Python version. Use whatever comes bundled with the `node` Docker image. All of the Kuma and Kumascript tests still pass.

I just wanted to get this work submitted, since I had done it a while ago and, with everything going on, didn't want to lose it.


callahad:
👍

PR #1295, opened 2019-12-20
Remove Scratchpad from Tools sidebar

jmswisher:


chrisdavidmills:
@jmswisher I have tried to test this locally but failed, as my Kuma install is now failing. I'll have to debug this with Ryan next year.

However, one immediate problem you've got here is that the travis build is failing — you can check out what the fail point is on the "Details" links.

The HTML structure looks OK, but I wonder if the test is failing because you've changed it to use the more expanded syntax form?

SebastianZ:
Chris is right. The test doesn't trim the values before comparing them with the expected ones. I think the test should be changed to compare the trimmed values.

Sebastian

jmswisher:
Thanks, VS Code, for the unhelpful reformatting.

PR #1297, opened 2019-12-23
Adds CSS Overflow Level 4 spec

rachelandrew:
While working on https://github.com/mdn/sprints/issues/2402 spotted that `text-overflow` had incorrect spec links as the definitions have moved to the CSS Overflow spec, adding the new Level 4 spec to correct that.


escattone:
@rachelandrew Wow, sorry this took so long! I don't normally review these, but was scanning for an obvious one and spotted this. Hope all is well.

PR #1300, opened 2019-12-24
Drop obsolete "Pointer Events 2 Ext" from SpecData

sideshowbarker:
The https://w3c.github.io/pointerevents/extension.html spec is now 404; https://github.com/w3c/pointerevents/commit/add40d9 removed it from the W3C Pointer Events repo. (The content has been incorporated directly into the Pointer Events spec.)

### DIFF
--- a/macros/SpecData.json
+++ b/macros/SpecData.json
@@ -1099,11 +1099,6 @@
     "url": "https://www.w3.org/TR/pointerevents2/",
     "status": "REC"
   },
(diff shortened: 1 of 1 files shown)


sideshowbarker:
FYI: In https://wiki.developer.mozilla.org/en-US/docs/Web/API/PointerEvent$compare?to=1596778&from=1583424 I removed the reference to this spec from the MDN PointerEvent article.

PR #1301, opened 2019-12-24
Add CSS4 Overflow to SpecData

sideshowbarker:
https://wiki.developer.mozilla.org/en-US/docs/Web/CSS/text-overflow has a Specifications table with a macro which references "CSS4 Overflow", but the MDN backend doesn’t (yet) recognize that value, so it displays "Unknown".


sideshowbarker:
oops I see now this is a dupe of https://github.com/mdn/kumascript/pull/1297

PR #1302, opened 2019-12-24
Add CSS Properties and Values API to SpecData

sideshowbarker:
https://wiki.developer.mozilla.org/en-US/docs/Web/API/CSS/RegisterProperty has a Specifications table with a macro which references "CSS Properties and Values API", but the MDN backend doesn’t (yet) recognize that value, so it displays "Unknown".


sideshowbarker:
Cheers & thanks :)

PR #1304, opened 2019-12-27
Use Node 12

### LINKED ISSUE #1303: Why Node 10 when kuma uses Node 12?
This project's [Dockerfile uses Node 10](https://github.com/mdn/kumascript/blob/fde760593d47ab323db5e1a2f18b2ff9f7902861/docker/Dockerfile#L1)
but [in Kuma we're using Node 12](https://github.com/mdn/kuma/blob/865deab60d6a97966e13eff6f1bbf492e1631afe/docker/images/kuma_base/Dockerfile#L4)

Perhaps I'll never understand how kuma and kumascript connect and relate in terms of docker and releases but shouldn't it make sense to CI Kumascript with the same version of Node as what we're going to be using in Kuma? @escattone 

Also, Node 10 is *not* the current LTS version.

peterbe:
Fixes #1303

### DIFF
--- a/docker/Dockerfile
+++ b/docker/Dockerfile
@@ -1,4 +1,4 @@
-FROM node:10.18.0@sha256:15a23eefad1798c98238fe71b22b65e20c26c17f139505cf6abcab53a740554a
+FROM node:12.14.0
 
 RUN set -ex && \
     apt-get update && \
--- a/.travis.yml
+++ b/.travis.yml
@@ -2,7 +2,7 @@ sudo: required
 
 language: node_js
 node_js:
-    - "10"
+    - "12"
 
 services:
   - docker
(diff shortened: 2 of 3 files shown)


docker/Dockerfile
```diff
@@ -1,4 +1,4 @@
-FROM node:10.18.0@sha256:15a23eefad1798c98238fe71b22b65e20c26c17f139505cf6abcab53a740554a
+FROM node:12.14.0
```
peterbe: Renovate will add the digest once this is in master.

peterbe:
Are there any other places where the Node version is mentioned?

Is it right that we're building KS with Node 12 in kuma? But as a standalone project, KS uses Node 10?

PR #1313, opened 2020-01-17
Add `tfoot` (for less user scroll-bouncing)

Zearin:
I wanted to put the row containing the totals in the `<tfoot>` as well, but I couldn’t think of a way to do it without rewriting the whole loop.

At least this way, the totals remain at the bottom, and they will have the `<tfoot>` directly beneath them to give the reader the columns for each total.


peterbe:
@ExE-Boss Can you test this one?

@Zearin Also, if you can, it would be really helpful to see some screenshots of what this PR changes and perhaps some specific pages where your new feature would be really useful.

PR #1317, opened 2020-01-22
Remove obsolete {{jsindex}} macro

wbamberg:
See https://github.com/mdn/stumptown-content/issues/274 for some context, especially https://github.com/mdn/stumptown-content/issues/274#issuecomment-576468538 and the subsequent comments.

This macro was used in 2 en-US pages (method and property index pages) and 24 pages in all locales. I've tried to change the pages that used this macro into redirects to JavaScript/Reference for all locales. It's hard to verify though because https://github.com/mdn/kuma/issues/6362 means the calls are still listed in our search results:

https://wiki.developer.mozilla.org/en-US/search?locale=*&kumascript_macros=jsindex&topic=none&per_page=100

But you can verify by clicking through the search results.

This macro is not called internally either:

https://github.com/mdn/kumascript/search?q=jsindex&unscoped_q=jsindex


peterbe:
Oh. You've already taken care of all the other 24. From the prod DB:

```
>>> Document.deleted_objects.filter(html__contains='jsindex').count()
0
```

wbamberg:
Thanks Peter!

PR #1321, opened 2020-01-27
Add CSS Easing Functions spec to SpecData

sideshowbarker:
https://github.com/w3c/csswg-drafts/commit/322502c renamed the "CSS Timing Functions" spec to "CSS Easing Functions".  https://drafts.csswg.org/css-timing-1/ now redirects to https://drafts.csswg.org/css-easing-1/

### DIFF
--- a/macros/SpecData.json
+++ b/macros/SpecData.json
@@ -214,6 +214,11 @@
     "url": "https://drafts.csswg.org/css-display/",
     "status": "CR"
   },
+  "CSS Easing 1": {
+    "name": "CSS Easing Functions Level&nbsp;1",
+    "url": "https://drafts.csswg.org/css-easing-1/",
+    "status": "CR"
+  },
   "CSS3 Environment Variables": {
     "name": "CSS Environment Variables Module Level&nbsp;1",
     "url": "https://drafts.csswg.org/css-env-1/",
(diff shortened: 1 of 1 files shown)


escattone:
@sideshowbarker So sorry for the delayed response. I'm trying to chip away at some of these as I have time. Since you're adding the `CSS Easing 1` entry, I'm wondering if the following entry it replaces should be removed?
```json
  "CSS Timing 1": {
    "name": "CSS Timing Functions Level&nbsp;1",
    "url": "https://drafts.csswg.org/css-timing-1/",
    "status": "WD"
  },
```
cc @chrisdavidmills

sideshowbarker:
> Since you're adding the `CSS Easing 1` entry, I'm wondering if the following entry [`CSS Timing 1`] it replaces should be removed?

Yep — thanks, I’ve amended the commit to remove that

https://developer.mozilla.org/en-US/docs/Web/CSS/easing-function is the only place in MDN where the `CSS Timing 1` macro is used, so after this change gets merged and deployed, I’ll update that MDN article to use the `CSS Easing 1` macro instead.

escattone:
@sideshowbarker 

> Yep — thanks, I’ve amended the commit to remove that
>
> https://developer.mozilla.org/en-US/docs/Web/CSS/easing-function is the only place in MDN where the `CSS Timing 1` macro is used, so after this change gets merged and deployed, I’ll update that MDN article to use the `CSS Easing 1` macro instead.

Thanks!

escattone (approved):

PR #1335, opened 2020-03-03
dont double escape page titles

peterbe:
Fixes https://github.com/mdn/kuma/issues/6475

Screen Shot 2020-03-03 at 5 17 01 PM

I'm a little bit uncertain about what's going on here but the change works. I've tried it in Kuma. 
If you make this change, you get that above screenshot. 
Seems the [macros hasn't been touched in over a year](https://github.com/mdn/kumascript/blob/master/macros/Index.ejs)

But if you look at [this line](https://github.com/mdn/kumascript/blob/2e04c8d31bfb6900a82cd69cc0fd3e6110ed307d/macros/Index.ejs#L49) it builds up an array of pages. And for each title it transforms `<tag>` to `&lt;tag&gt;` with manual string replace. So, any HTML turns into already escaped HTML. Then, when you've done that you can confidently use `<%-` instead of `<%=`. 

However, it used to do 
```javascript
... title.replace('<','&lt;').replace('>','&gt;')
```
which is dangerous because of this example:
```
> var title = '<tag>'
'<tag>'
> title.replace('<','&lt;').replace('>','&gt;')
'&lt;tag&gt;'
> title += ' <other>'
'<tag> <other>'
> title.replace('<','&lt;').replace('>','&gt;')
'&lt;tag&gt; <other>'
```
If you don't use `/REGEX/g` it won't replace EVERY occurance. But this works:
```
> title.replace(/</g,'&lt;').replace(/>/g,'&gt;')
'&lt;tag&gt; &lt;other&gt;'
```

### DIFF
--- a/macros/Index.ejs
+++ b/macros/Index.ejs
@@ -68,7 +68,7 @@ getPages(pageList);
 <% for (i=0; i < result.length; i++) { %>
 <tr>
     <td rowspan="2"><%=i+1%></td>
-    <td rowspan="2"><a href='<%=result[i].url%>'><%=result[i].title%></a></td>
+    <td rowspan="2"><a href='<%=result[i].url%>'><%-result[i].title%></a></td>
     <td><strong><%=result[i].tags.join(", ")%></strong></td>
 </tr>
 <tr>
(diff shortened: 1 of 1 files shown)


peterbe:
Note! I don't know how to write and run unit tests in kumascript.

PR #1339, opened 2020-03-06
Add Web Animations Level 2 spec

lpd-au:
There is now a draft spec for Web Animations Level 2, it can be found here: https://drafts.csswg.org/web-animations-2/

Some features already documented on MDN have been moved from Level 1 to Level 2, for example: https://developer.mozilla.org/en-US/docs/Web/API/KeyframeEffect/iterationComposite


lpd-au:
@chrisdavidmills you may be a good reviewer for this.

PR #1348, opened 2020-03-22
Community link updates

rebloor:
Amended structure and content of community/channels content based on [this request](https://docs.google.com/document/d/1Rz2J0m1KGQOs6-BFvnSfhIuUJFWKc01XrAbHiv9Mbhc/edit?usp=sharing).

Please note that I am unable to run Kuma locally, so haven't previewed/tested these changes.


peterbe:
@rebloor Are you able to run the tests locally? That seems to be wha's breaking in CI.

rebloor:
@peterbe 

> Please note that I am unable to run Kuma locally, so haven't previewed/tested these changes.

Unfortunately, attempts to set up Kuma locally fail because the clone of the subproject `locale` does not finish. 

I have looked at the build log output and hopefully have identified the cause of the problem.

peterbe:
Ah. So you can't run the tests on `kumascript` without a full `kuma`. Sigh.

rebloor:
@wbamberg  Can you offer any advice or suggest the right person to talk to?  Unfortunately, I'm unable to get kuma preview running so cannot check any of this locally. The build is giving me an error `"Error: Bad value “https://chat.mozilla.org/#/room/#addons:mozilla.org” for attribute “href” on element “a”: Illegal character in fragment: “#” is not allowed."` from a  line 564: `<li><a href="https://chat.mozilla.org/#/room/#addons:mozilla.org">Matrix</a></li>`

wbamberg:
We have a test for the `{{AddonSidebar}}` macro that it emits valid HTML, and the validator is complaining that "https://chat.mozilla.org/#/room/#addons:mozilla.org" isn't a valid URL, I think because the first "#" is treated as a fragment identifier, and then fragment identifiers aren't themselves allowed to contain "#".

We use the W3 validation checker for this, and you can see the same error by pasting something like this:

```html
<!DOCTYPE html>
<html lang="en">
  <head>
    <meta charset="utf-8">
    <title>example page</title>
  </head>
  <body>
    <a href="https://chat.mozilla.org/#/room/#addons:mozilla.org">blah</a>
  </body>
</html>
```

...into https://validator.w3.org/#validate_by_input.

I've no reason to doubt the validator here and a read of https://www.ietf.org/rfc/rfc3986.txt seems to confirm that [the fragment should not contain "#"](https://www.ietf.org/rfc/rfc3986.html#section-3.5).

I think one option would be to encode the second "#", like: "https://chat.mozilla.org/#/room/%23addons:mozilla.org". This seems to pass the validator, and to work as a link to the addons channel.

> Please note that I am unable to run Kuma locally, so haven't previewed/tested these changes.

I think if you're making a substantial number of changes to KS macros, you should have Kuma running locally so you can test them first. I'm sympathetic that this is an annoying and time-consuming process, but not checking changes before asking for review is effectively moving that burden onto reviewers. I think it is OK to ask for help in #mdn-dev to get things working.

rebloor:
@wbamberg 

Ironically, I noticed your email just before I went to shut my chickens in for the night and while outside thought to myself "escape the URL" followed closely by "I bet that's what Will's email will suggest". And, as I write this I see that the checks have passed. Thanks

> I think if you're making a substantial number of changes to KS macros, you should have Kuma running locally so you can test them first. I'm sympathetic that this is an annoying and time-consuming process, but not checking changes before asking for review is effectively moving that burden onto reviewers. I think it is OK to ask for help in #mdn-dev to get things working.

I have been chasing a solution since last year, including asking for help in #mdn-dev, but no one seems be able to figure out what's wrong. In summary, the issue appears to come down to "clone of 'git://github.com/mozilla-l10n/mdn-l10n.git' into submodule path '/Users/sherpa/Documents/repositories/mdn/kuma/locale' failed".

wbamberg:
Chickens, is there anything they can't do? :)

rebloor:
@wbamberg As the Travis build has passed, can this be approved?

wbamberg:
I've tested this locally and it works in en-US. However:

* not my call, but IMO "Matrix" is not a very comprehensible name for the link. "Chat" (or "Addons chat" to go alongside "Add-ons blog" and "Add-on forum") might be better.
* the "Contact us" link won't work outside en-US. For example, on a page like https://developer.mozilla.org/de/docs/Mozilla/Add-ons/WebExtensions it will take you to https://developer.mozilla.org/de/docs/Mozilla/Add-ons#Contact_us, but that page doesn't have an element with a "Contact_us" id (if it even has a translation of the "Contact us" section, it's likely to get an id that's the German translation of "Contact us").

rebloor:
Thanks for the feedback @wbamberg  before I make any changes, can you think of any practical reason for not splitting out the "Contact us" details onto separate page?

wbamberg:
> can you think of any practical reason for not splitting out the "Contact us" details onto separate page?

I can't think of any. But it might still be a good idea to keep a pointer to the new page from https://developer.mozilla.org/en-US/docs/Mozilla/Add-ons#Contact_us, in case other things are pointing there.

rebloor:
@wbamberg I have made the changes per your suggestion, could you take another look?

wbamberg:
@rebloor , this PR has a merge conflict, can you please resolve that before I review?

rebloor:
@wbamberg conflict resolved, all yours

wbamberg:
This works fine, thank you for the updates! Do you think this reads weirdly:

```
Channels
    Add-ons blog
    Add-on forums
    Add-on chat
```

Should it be "Add-ons" throughout? "Add-ons forums" is awful, but maybe "Add-ons forum" would be OK?

rebloor:
@wbamberg Sorry, too busy worrying about structure :-( fixed

Issue #740, opened 2018-07-13
Drop "canonical order" in CSS Properties template


dontcallmedom:
IF I understand this right, the information about the canonical order of the grammar of CSS properties is only useful to browser implementors, not to Web developers (at least not to the vast majority of Web developers).

I think it would make more sense NOT to include that information.

[example of this info raising more questions than answering them](https://stackoverflow.com/questions/28963536/what-does-canonical-order-mean-with-respect-to-css-properties)

dontcallmedom:
yes, I believe this would mean editing cssinfo.ejs, and probably remove the relevant data from mdn/data.

Regarding additional improvements to the box, I think the "Media" field is also of limited value, and one from which the [CSS WG seems to be moving away from](https://github.com/w3c/csswg-drafts/issues/2413#issuecomment-402374797).

For other fields, it would probably be interesting to user-test them to see if they're understandable / useful - my hunch is that they could be more clearly explained, and maybe presented in a more graphical (or least more structured) way.

wbamberg:
I thought this could use a bit more discussion, so I started a [thread about it in Discourse](https://discourse.mozilla.org/t/updates-to-the-css-info-boxes/30855/). It seems like we have something like consensus for removing "Canonical order" and "Media".

SebastianZ:
To copy the statement from [my comment on the Discourse thread](https://discourse.mozilla.org/t/updates-to-the-css-info-boxes/30855/7?u=sebastianzartner), yes, let's remove "Media" and "Canonical order". For the vast majority of properties they don't provide any relevant info. And the cases where they do provide some info can be/are already covered by the textual description of their values.

Sebastian

wbamberg:
Almost 2 years later, this is now done.

I've merged https://github.com/mdn/kumascript/pull/1356, Ryan made a new deployment today and I've refreshed all 1007-odd CSS pages, and "Media" and "Canonical order" are both gone from these boxes.

I'd still like to remove the underlying data from mdn/data, but that's another story.

PR #1356, opened 2020-04-15
Remove cssinfo media and canonical order

wbamberg:
Fix for https://github.com/mdn/sprints/issues/2880, and also https://github.com/mdn/kumascript/issues/740.

I started looking into this because of https://github.com/mdn/sprints/issues/2880, which reports that the "Media" entry in CSS info boxes is broken sometimes. Since [we agreed ages ago to retire this item and also "Canonical order"](https://discourse.mozilla.org/t/updates-to-the-css-info-boxes/30855), just removing them seemed the best way to fix the bug.

Initially I thought this might require us to remove the underlying data, but having looked at the CSSInfo macro it seems that just not displaying these items is simple. I've tested this in a local Kuma and it seems to do the right thing.

Still, this macro doesn't have tests so a careful review is needed.

It would still be nice to:
(1) understand which items in mdn/data are actually used by third parties
(2) understand which CSS info box items we actually want to display

...but this seems like a step in the right direction, anyway.

### DIFF
--- a/macros/CSSInfo.ejs
+++ b/macros/CSSInfo.ejs
@@ -293,10 +293,6 @@ if (name === "preview-wiki-content") {
     }
 
     properties = properties.concat({
-        name: "media",
-        label: localize(localStrings, "media")
-    },
-    {
         name: "computed",
         label: await template("Xref_csscomputed")
     });
@@ -307,11 +303,6 @@ if (name === "preview-wiki-content") {
             label: localize(localStrings, "animationType")
         });
     }
-
-    properties = properties.concat({
-        name: "order",
-        label: localize(localStrings, "canonicalOrder")
-    });
 
     if (cssInfo.stacking) {
         properties = properties.concat({


chrisdavidmills:
Apologies for the delay on this @wbamberg ; I didn't get an email confirmation about it. @mention me in future, just to be sure!

So, this looks like a sensible way to fix this. Unfortunately, I can't currently test it because my local Kuma install has gone wrong again, and I don't how to fix it. It won't let me log in, and isn't rendering any KS macros. I'll bring this up with @escattone when I chat to him today.

wbamberg:
> Apologies for the delay on this @wbamberg ; I didn't get an email confirmation about it. @mention me in future, just to be sure!

No worries! I am missing email notifications sometimes as well. I think maybe something has changed with GitHub notifications.

> So, this looks like a sensible way to fix this. Unfortunately, I can't currently test it because my local Kuma install has gone wrong again, and I don't how to fix it. It won't let me log in, and isn't rendering any KS macros. I'll bring this up with @escattone when I chat to him today.

The way I do this now I don't create a local Kuma account. I just do something like:

```
docker exec kuma_worker_1 python ./manage.py scrape_document https://developer.mozilla.org/en-US/docs/Web/CSS/transform
docker-compose exec web ./manage.py render_document /en-US/docs/Web/CSS/transform
```

...the first line gets a new document containing the macro, the second one renders the KS macros. If I then change the KS macros, I need to do `docker-comppose restart`, then run the `render-document` line again to see the results.

chrisdavidmills:
@wbamberg Brilliant, this worked great, and my local Kuma is now rendering KS macros again.

I've checked it out, and your update works fine.

PR #1357, opened 2020-04-16
adding new tooling articles to the learning area sidebar

chrisdavidmills:
This PR adds the following to the Learning Area sidebar:

* https://developer.mozilla.org/en-US/docs/Learn/Tools_and_testing/GitHub
* https://developer.mozilla.org/en-US/docs/Learn/Tools_and_testing/Understanding_client-side_tools, and all of its sub-srticles

### DIFF
--- a/macros/LearnSidebar.ejs
+++ b/macros/LearnSidebar.ejs
@@ -179,6 +179,7 @@ var text = mdn.localStringMap({
         'Introduction_to_automated_testing' : 'Introduction to automated testing',
         'Setting_up_your_own_test_automation_environment' : 'Setting up your own test automation environment',
       'Git_and_GitHub' : 'Git and GitHub',
+        'Git_and_GitHub_overview' : 'Git and GitHub overview',
       'Understanding_client-side_web_development_tools' : 'Understanding client-side web development tools',
         'Client-side_web_development_tools_index' : 'Client-side web development tools index',
         'Client-side_tooling_overview' : 'Client-side tooling overview',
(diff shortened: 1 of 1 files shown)


chrisdavidmills:
I've tested this on my local Kuma install and it looks to be working fine. @Elchi3 , can you give it a quick verification?

macros/LearnSidebar.ejs
```diff
@@ -1724,6 +1780,20 @@ var text = mdn.localStringMap({
         </ol>
     </details>
   </li>
+  <li><a href="<%=baseURL%>Tools_and_testing/GitHub"><%=text['Git_and_GitHub']%></a></li>
```
Elchi3, line 1783: It works for me, too, but the "Git and GitHub" link is now the only one without sub pages and thus it looks a bit lost to me.

learnsidebar

chrisdavidmills: Yeah, I agree with you, but I wasn't sure what to do about that. I mean, it is a page at the module level, but it doesn't link to articles on our site — it links to external GitHub articles.

Maybe I could just remove it.

Elchi3: Or you add the guides listed here? http://wiki.localhost.org:8000/en-US/docs/Learn/Tools_and_testing/GitHub#Guides

chrisdavidmills: OK, I've just tried adding them. I've not added in language strings for the external articles, as they are not translated. I'll try testing the update now.

chrisdavidmills: Dammnit, I can't seem to test it. I pulled the changes to my local kumascript submodule, then restarted docker, but it won't show me the updates — it just shows me the original sidebar update as it was at the start of this PR....

macros/LearnSidebar.ejs
```diff
@@ -1724,7 +1787,33 @@ var text = mdn.localStringMap({
         </ol>
     </details>
   </li>
-
+  <li><a href="<%=baseURL%>Tools_and_testing/GitHub"><%=text['Git_and_GitHub']%></a></li>
```
Elchi3, line 1790: I tested it and it works! Please remove this first list element though:

```suggestion
```

chrisdavidmills: Oops; I left that in by mistake! Removed now.

PR #1362, opened 2020-04-22
Void the SeeCompatTable macro

### LINKED ISSUE #1360: Make SeeCompatTable macro do nothing
Since we don't want the SeeCompatTable banner at the tops of pages anymore, and we will be importing all this content into another platform at some point anyway, let's make that macro a no-op so we don't have to actually go and remove that from everyplace currently using it.

peterbe:
part of #1360

Before:
Screen Shot 2020-04-22 at 12 34 49 PM

After:
Screen Shot 2020-04-22 at 12 34 57 PM

### DIFF
--- a/macros/SeeCompatTable.ejs
+++ b/macros/SeeCompatTable.ejs
@@ -1,17 +1,6 @@
 <%
-
-var str = mdn.localString({
-    "de"    : "<strong>Dies ist eine experimentelle Technologie</strong><br />Da diese Technologie noch nicht definitiv implementiert wurde, sollte die <a href='#Browser_compatibility'>Browserkompatibilität</a> beachtet werden. Es ist auch möglich, dass die Syntax in einer späteren Spezifikation noch geändert wird.",
-    "es"    : "<strong>Esta es una <a href='/en-US/docs/MDN/Contribute/Guidelines/Conventions_definitions#Experimental'>tecnología experimental</a></strong><br />Comprueba la <a href='#Browser_compatibility'>Tabla de compabilidad de navegadores</a> cuidadosamente antes de usarla en producción.",
-    "fr"    : "<strong>Cette fonction est expérimentale</strong><br />Puisque cette fonction est toujours en développement dans certains navigateurs, veuillez consulter le <a href='#Compatibilité_des_navigateurs'>tableau de compatibilité</a> pour les préfixes à utiliser selon les navigateurs.<br />Il convient de noter qu'une fonctionnalité expérimentale peut voir sa syntaxe ou son comportement modifié dans le futur en fonction des évolutions de la spécification.",
-    "ja"    : "<strong>これは<a href='/ja/docs/MDN/Contribute/Guidelines/Conventions_definitions#Experimental'>実験的な機能</a>です。</strong>本番で使用する前に<a href='#Browser_compatibility'>ブラウザー実装状況</a>をチェックしてください。",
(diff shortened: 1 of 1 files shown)


peterbe:
@escattone watch out to re-clone this into the ks-renderer soon.

Elchi3:
Is there a reason we are doing this now out of a sudden? I agree to the removal per se, just wondering.
Also, with Ryan's revise tool, we have actually removed macro calls from wiki content and then removed the macro code from the ks repo. Why is a no op and leaving the cruft in content better in this situation?

peterbe:
> Is there a reason we are doing this now out of a sudden? I agree to the removal per se, just wondering.

Can't speak to that. 

> Also, with Ryan's revise tool, we have actually removed macro calls from wiki content and then removed the macro code from the ks repo. Why is a no op and leaving the cruft in content better in this situation?

No good answer but the act of turning it into a "void macro" was so easy. And the day we've moved to git, we can do `grep -r SeeCompatTable content | xargs sed /{{SeeCompatTable}}//s` and quickly fix all of them in one quick sweep :)

Elchi3:
We've been quite carefully planning when to render/remove/replace macros as part of content cleaning. If that carefulness doesn't apply here, I'd like to know why. For the writing experience, ripping out the macro call from content seems better to me and I guess using the revise tool is just as straight forward as using grep on day X. In https://github.com/mdn/sprints/issues/2524 we have done so successfully without any problem for other macros.

peterbe:
> We've been quite carefully planning when to render/remove/replace macros as part of content cleaning. If that carefulness doesn't apply here, I'd like to know why. For the writing experience, ripping out the macro call from content seems better to me and I guess using the revise tool is just as straight forward as using grep on day X. In [mdn/sprints#2524](https://github.com/mdn/sprints/issues/2524) we have done so successfully without any problem for other macros.

If using the revise tool to fix 3.4k pages is easy, then I think we should opt for that instead. But I would argue there's no point in discarding this PR since it was easy to do and already done :)

But what I will do is change this PR's description so it doesn't *close* the issue. Using the revise tool to actually do the cleanup would be better and its effort can be the one that closes the issue. Is it hard to run it? (CC @escattone )

a2sheppy:
I was under the impression that the plan had always been to turn the macro into a no-op in the interim until we finished removing it from the pages. Certainly, this is a way to get rid of it throughout the content without having to do the editing right off. But if the plan has changed, then mea culpa.

Certainly there’s no urgency. It’s just something that’s been driving me nuts lately; I keep finding brand-new pages that people are still using this macro on, and if it didn’t do anything, perhaps they’d be more likely to give up on using it — and at least it wouldn’t be causing that big yellow blob to appear any more.


Eric "Sheppy" Shepherd
Senior Technical Writer
MDN Web Docs
Mozilla

wbamberg:
We recently spent some time triaging KS macros in the JavaScript docs, and explicitly discussed this class of macro ("banners" or "headers"), and why we didn't feel able to remove them at this point, here: https://github.com/mdn/sprints/issues/2520#issuecomment-581734033 .

If this reasoning is wrong it would be great to have that conversation over there, so we can remember it.

If we ever get out of HTML+KS-soup-land, we would like the renderer to display banners based on compat data, and that was the point of https://github.com/mdn/stumptown-content/issues/306 and https://github.com/mdn/stumptown-content/pull/354.

PR #1363, opened 2020-04-26
IDB 3 now has new things not in IDB 2

jyasskin:
For example, https://wiki.developer.mozilla.org/en-US/docs/Web/API/IDBFactory/databases claims to be a REC, but it's not in https://www.w3.org/TR/IndexedDB/.


chrisdavidmills (approved):
LGTM, r+

PR #1369, opened 2020-05-08
Mark WebAuthn spec as REC

sideshowbarker:
Fixes https://github.com/mdn/sprints/issues/3229


equalsJeffH:
perhaps this MDN machinery could automagically extract specs' maturity level?

sideshowbarker:
> perhaps this MDN machinery could automagically extract specs' maturity level?

I think it’s more likely no indicators of WD/CR/PR/Rec status will be shown at all in the next MDN evolution. At least my sense of it is, the prevailing agreement at this point leans toward concluding that the WD/CR/PR/Rec status isn’t useful information for web developers..

PR #1403, opened 2020-07-07
adding Applicationspanel to sidebar

chrisdavidmills:
Adding the DevTools Application panel documentation to the tools sidebar, which is going to live at https://developer.mozilla.org/en-US/docs/Tools/Application.


peterbe:
@escattone If you merges, do we need to copy it over to yari again?

PR #1405, opened 2020-07-14
add browser release date to compatibility table

joelhawksley:
### Problem statement
As a developer looking at a feature on MDN to check its browser compatibility, the browser version number data is not super meaningful to me, because I don't know the release date for each browser version.

### Proposed solution
This PR adds a `title` attribute to each version number cell in the browser compatibility table with the release date for the browser version.

- Joel & Anonymous developer Joel is mentoring

### DIFF
--- a/macros/Compat.ejs
+++ b/macros/Compat.ejs
@@ -228,6 +228,31 @@ function getCellString(added, removed, partial) {
 }
 
 /*
+Returns the version number to appear in a cell
+
+`added` and `removed` are either null, true, false or a string containing a version number
+`partial` is either null, true, or false indicating partial_implementation
+*/
+function getVersionNumber(added, removed, partial) {
+  let output = undefined;
+
+  if(added !== null && added !== true && added !== false) {
+    output = added;
+  }
+
+  if (removed) {
+    output = undefined;
+  } else if (partial) {
+    output = undefined;
+    // Display "Partial" instead of "Yes", "No", or "?" if we have no version string
+    if (!(typeof(added) !== 'string')) {
+      output = added;
+    }
+  }
+  return output;
+}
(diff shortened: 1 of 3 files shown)


peterbe:
Thanks @joelhawksley ! I like the feature suggestion a lot. Actually, it's a piece of information that I too often wonder. Especially in browsers I'm not super familiar with. (I work at Mozilla so I'm very familiar with what it means to say "Firefox 79" for example). 

The implementation of the BCD table has actually be rewritten, from the ground up, in a new project we call [Yari](https://github.com/mdn/yari) which we soon (many months realistically) will replace `kuma` and `kumascript` to generate MDN pages. 

That new repo is actually very easy to work with. We separate the code that extracts the structured data away (a JavaScript Node build script) from the code that "paints" the BCD table (a React component in TypeScript). 

Because I was so curious in saying a comment with more intimate knowledge about how feasible it would be; I just started a prototype that does it. First I filed your "feature request" for you here: https://github.com/mdn/yari/issues/943 
Then I started hacking on a quick prototype to prove that it can be done and clearly it can:
Screen Shot 2020-07-20 at 4 34 24 PM

If you're interested, I can tidy up my hack, and together with my fellow core-contributor, perhaps you can help out by checking out the branch and the PR for some feedback.

peterbe:
FYI, here's a suggestion to fix it in Yari: https://github.com/mdn/yari/pull/944

joelhawksley:
@peterbe fantastic! 

We weren't sure whether to submit a patch here or to Yari, given that Yari is not yet in production. 

Thanks for writing the implementation in Yari! I'm sure many folks will find this behavior helpful when it goes live ❤️

PR #1406, opened 2020-07-14
adding tooling module to sidebar

chrisdavidmills:
I completely failed to actually add [our new tooling module](https://developer.mozilla.org/en-US/docs/Learn/Tools_and_testing/Understanding_client-side_tools) to the learn sidebar, so this PR does that.

Plus I've reordered the tools and testing modules so that they appear in the same order as they do in the topic page: https://developer.mozilla.org/en-US/docs/Learn/Tools_and_testing.

@Elchi3 can you give this a quick check to see if you think it makes sense?

Then @escattone can you please refresh the /Learn pages once this is done?

This is not hyper urgent, but it'd be nice to get it done before too much longer has passed.


Elchi3:
Collapsed: 
temp0

<hr>

Expanded:

temp1

Elchi3:
How about removing "Understanding" so that the line isn't too long?

chrisdavidmills:
@Elchi3 good point; done!

PR #1422, opened 2020-09-22
CSS Containment spec. updates

SebastianZ:
The URL for the CSS Containment specification didn't contain a version number. Because the spec. already advanced to version 2 now, the URL linked to the new version instead of the old one. Therefore I have added the version 1 now.

Additionally, I have added the second level of the specification.


escattone (approved):
Thank you @SebastianZ!

PR #1427, opened 2020-10-07
use new @mdn/browser-compat-data

### LINKED ISSUE #1425: use @mdn/browser-compat-data rather than deprecated mdn-browser-compat-data
See https://github.com/mdn/browser-compat-data/issues/6640

escattone:
Fixes #1425 

Per https://github.com/mdn/browser-compat-data/issues/6640, stop using deprecated `mdn-browser-compat-data` npm package and use the new `@mdn/browser-compat-data` package instead.

### DIFF
--- a/docker/Dockerfile
+++ b/docker/Dockerfile
@@ -22,7 +22,7 @@ WORKDIR /
 COPY kumascript/package.json kumascript/npm-shrinkwrap.json /
 RUN npm ci && \
     # update any top-level npm packages listed in package.json,
-    # such as mdn-browser-compat-data,
+    # such as @mdn/browser-compat-data,
     # as allowed by each package's given "semver".
     npm update
 ENV NODE_PATH=/node_modules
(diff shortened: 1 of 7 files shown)


peterbe (approved):
I'm not entirely sure how to test this. 
I did `cd kumascript && git remote escattone ... && git checkout escattone/use-new.. && cd ..` and then a full `docker-compose pull`. 
But before I started, just make sure it's using these files and something from the image, I put in a `console.log(...)` in the `kumascript/macros/Compat.ejs` but it never prints anywhere in `docker-compose up`. 
So I can't tell if it's using the local `./kumascript/` or if it's using the DockerHub image. Sigh. 

But each individual line of change looks correct and very sane. 
I guess we get a chance to try this again before we make a prod release. Just to be safe.

escattone:
@peterbe Just FYI, I built and ran (via `docker-compose`) this kumascript image locally, double-checked that I could `require("@mdn/browser-compat-data")`, and then re-rendered a page with the `compat` macro. All of it was successful.

PR #1428, opened 2020-10-11
Make JSRef not claim Object inherits from Function

sideshowbarker:
This change makes the code for the JSRef macro not set the inheritance data for Object to Function.

Otherwise, without this change, when viewing the MDN article at https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Global_Objects/Object, the sidebar there incorrectly shows Object inheriting from Function.


peterbe:
@escattone @chrisdavidmills If this is a valid diff, we have to remember to manually copy this over to Yari after merge.

escattone (approved):
Good catch, thanks @sideshowbarker!

PR #1431, opened 2020-10-19
add MathML Core spec

rachelandrew:
While documenting for this bug https://bugzilla.mozilla.org/show_bug.cgi?id=1665975 I am adding a page for `math-style` which is defined in the MathML Core spec: https://mathml-refresh.github.io/mathml-core/#the-math-style-property

This PR adds that spec.


escattone (approved):
Thanks @rachelandrew !

PR #1440, opened 2020-11-15
Both AudioContexts inherit BaseAudioContext

MendyBerger:
According the [the spec](https://www.w3.org/TR/webaudio/#BaseAudioContext) both AudioContext and OfflineAudioContext inherit from BaseAudioContext.

This would probably also resolve mdn/sprints/issues/3565


chrisdavidmills (approved):
Yup, looks correct to me. Thanks!

PR #1447, opened 2020-12-12
File System Access API

Rumyra:
Adding File System Access API 

https://wicg.github.io/file-system-access/


escattone (approved):
Thanks @Rumyra !

PR #1448, opened 2020-12-14
update the README to indicate archival

escattone:
Update README to indicate that this repository will be archived.


chrisdavidmills (approved):
The warning notice looks good, thanks @escattone .

I did wonder though — is the content below it even relevant anymore, and if so, should it be put inside a README in the next location and deleted from here?

escattone:
@chrisdavidmills 

> I did wonder though — is the content below it even relevant anymore, and if so, should it be put inside a README in the next location and deleted from here?

No, it's not relevant anymore at all.

chrisdavidmills:
> @chrisdavidmills
> 
> > I did wonder though — is the content below it even relevant anymore, and if so, should it be put inside a README in the next location and deleted from here?
> 
> No, it's not relevant anymore at all.

Maybe get rid of it then?

escattone:
> > @chrisdavidmills
> > > I did wonder though — is the content below it even relevant anymore, and if so, should it be put inside a README in the next location and deleted from here?
> > 
> > 
> > No, it's not relevant anymore at all.
> 
> Maybe get rid of it then?

Good point, will do!

